PR #18, opened 2020-06-05
Implement gossip-based node selection and discovery

### LINKED ISSUE #10: Add a cluster discovery mechanism
We need to be able to connect to an Event Store cluster. In the Event Store Dotnet Client we currently use the gossip endpoint which is http based. We are working on a gRPC based implementation.

jen20:
Fixes #10 

This pull request implements node selection for the gRPC client based on one or more seed nodes and a preference for connecting to a cluster leader, follower or read replica. As part of this it also implements a public client for the gossip service which can be used by library consumers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ import org.gradle.api.tasks.testing.logging.TestLogEvent
 plugins {
     id "java"
     id "com.google.protobuf" version "0.8.12"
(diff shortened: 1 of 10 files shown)


YoEight:
Done here: https://github.com/EventStore/EventStoreDB-Client-Java/pull/32

PR #32, opened 2020-08-26
Implement cluster connection.

YoEight:


ahjohannessen:
@YoEight FYI, https://github.com/grpc/grpc-java/issues/7133 - Here is an example of how I register a name resolver in my fs2 based [client](https://github.com/ahjohannessen/sec/blob/9ea98e1e63d79a9c4fc4d40f012a180fcd8d06de/core/src/main/scala/sec/api/cluster/watch.scala#L36).

YoEight:
Thanks for the pointer @ahjohannessen

YoEight:
I addressed the deprecation warning.

hayley-jean (approved):
Looks good 👍 
Only thing to note currently is that the client needs to catch `NotLeaderException`, and dispose and reconnect the connection to connect to the new leader.

PR #35, opened 2020-09-02
Implement Persistent Subscription API.

YoEight:


YoEight:
Waiting on #32 to handle `Not Leader` exception use cases.

YoEight:
Current status: Ack/Nack are currently ignored when sent. I'm using structured UUID constructor. Will try to use string constructor instead.

PR #49, opened 2020-12-22
Added limited projection functionality

rickybarefield:
Added functionality to create and get the result of a continuous projection.

A Projections API which provides functionality to create and get the result of a continuous projection.
Getting the result of a projection uses Jackson deserialization to avoid exposing the raw protobuf classes.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ThrowingFunction.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ThrowingFunction.java
@@ -1,0 +1,7 @@
+package com.eventstore.dbclient;
+
+@FunctionalInterface
+interface ThrowingFunction<TInput, TResult, TException extends Throwable> {
+
+    TResult apply(TInput first) throws TException;
+}
--- a/README.md
+++ b/README.md
@@ -90,6 +90,44 @@ public class Main {
 }
 ```
 
+## Projections
+
+This client currently supports creating and getting the result of a continuous projection.
+
+Create a projection:
+```java
+client.projections()
+    .createContinuous(PROJECTION_NAME, PROJECTION_JS, false)
+    .execute()
+    .get();
+```
+
+Define a class in which to deserialize the result:
+```java
+public class CountResult {
+
+    private int count;
+
+    public int getCount() {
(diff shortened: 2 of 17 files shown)


YoEight:
Thanks for your great work!

Could you squash your commits in the end?

db-client-java/src/main/java/com/eventstore/dbclient/ThrowingFunction.java
```diff
@@ -0,0 +1,7 @@
+package com.eventstore.dbclient;
+
+@FunctionalInterface
+interface ThrowingFunction<TInput, TResult, TException extends Throwable> {
```
YoEight, line 4: I fail to see the purpose of that interface, along with `ThrowingBiFunction`. Is this mostly for convenience?

rickybarefield: This is to allow lambda syntax where the expression throws a checked exception.  The standard functional interfaces don't allow for this in Java.  So for instance in GetProjectionResult, I've been able to use an instance of ThrowingBiFunction to hold the deserialization strategy which can throw a JsonProcessingException.

rickybarefield:
> Thanks for your great work!
> 
> Could you squash your commits in the end?

All done.

rickybarefield:
@YoEight - the test failure looks like an issue starting the test container? Can't imagine I've affected that with this commit?

YoEight:
> @YoEight - the test failure looks like an issue starting the test container? Can't imagine I've affected that with this commit?

Yes, testcontainers are flaky on GA, sometimes you end up with health-checking issues.

YoEight:
Tested on an EventStoreDB Cloud cluster and it works. Well done and thank you!

YoEight:
@rickybarefield do you need it released ASAP or can you wait?

rickybarefield:
> Tested on an EventStoreDB Cloud cluster and it works. Well done and thank you!

You're welcome

rickybarefield:
> @rickybarefield do you need it released ASAP or can you wait?

Nope, no rush for me, I can use one built locally.

PR #51, opened 2020-12-29
Enable signing only when publishing.

YoEight:
Before that patch, building the project on a machine not configured to publish on Maven Central resulted in that error:

```
> Task :db-client-java:signMavenJavaPublication FAILED
FAILURE: Build failed with an exception.
* What went wrong:
A problem was found with the configuration of task ':db-client-java:signMavenJavaPublication' (type 'Sign').
> No value has been specified for property 'signatory.keyId'.
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
```

This patch enables signing only if we want to publish.

### DIFF
--- a/db-client-java/build.gradle
+++ b/db-client-java/build.gradle
@@ -113,8 +113,11 @@ publishing {
         maven {
             url = 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
             credentials {
-                username = ossrhUsername
-                password = ossrhPassword
+                def publishLogin = project.hasProperty("ossrhUsername") ? project.getProperties().get("ossrhUsername") : ""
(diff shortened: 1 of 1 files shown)


oskardudycz (approved):
Sweet 😎

oskardudycz (changes requested):
After double-checking, it seems that's not working :|

It seems that the issue is with maven repository credentials, when I comment them out then build is passing.

YoEight:
Also fixed that maven credentials issue too.

oskardudycz (approved):
👍

PR #64, opened 2021-03-05
Improve EventData api

captainhorsepower:
- add EventDataBuilder#json method that accepts array of bytes (it's handy in my case)
- fix 'typo' in EventDataBuilder#binary

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
@@ -49,8 +49,8 @@ public final class EventData {
         return builderAsBinary(null, eventType, eventData);
     }
 
-    public static EventDataBuilder builderAsBinary(UUID id, String eventType, byte[] eventData) {
-        return EventDataBuilder.binary(eventType, eventData);
+    public static EventDataBuilder builderAsBinary(UUID eventId, String eventType, byte[] eventData) {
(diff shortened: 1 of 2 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
```diff
@@ -49,8 +49,8 @@ public static EventDataBuilder builderAsBinary(String eventType, byte[] eventDat
         return builderAsBinary(null, eventType, eventData);
     }
 
-    public static EventDataBuilder builderAsBinary(UUID id, String eventType, byte[] eventData) {
-        return EventDataBuilder.binary(eventType, eventData);
+    public static EventDataBuilder builderAsBinary(UUID eventId, String eventType, byte[] eventData) {
```
YoEight, line 52: This change is not needed.

YoEight (changes requested):
Looks good to me, thanks for your contribution!

db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
captainhorsepower: Do you mean renaming `id` -> `evetId` is not need or that `id` param was intentionally not passed to builder?

I case of renamimng, I just made it consistent with `#builderAsJson` method

YoEight (approved):

oskardudycz (approved):

PR #78, opened 2021-05-18
Support proper logging infrastructure.

### LINKED ISSUE #54: Increase significantly the log output.

YoEight:
Fixes #54


oskardudycz (approved):
@YoEight, I'm okay with those changes. I like that you chose a generic wrapper, so it's not limiting our users 👍 
Could you add some test if logging works as expected? :)

Lougarou (approved):
Tested with slf4j-simple and seems to be working fine.

PR #79, opened 2021-06-07
Support handling of large event streams.

### LINKED ISSUE #70: Improve working with a large set of events
Hi again :)

From what I understand when reading the [documentation](https://developers.eventstore.com/clients/grpc/reading-events/reading-from-the-all-stream.html#reading-forwards), reading from the "$all" stream returns a `List<ResolvedEvent>`. But returning all events in the entire eventstore as a List would be highly inefficient and will most likely lead to OOM. The same docs says "This will return an AsyncEnumerable that can be iterated on", see image:

image

but this is cannot be true in Java right? At least I've never come across such a thing in Java yet :) 

Is there another way to read all events in the all stream? If so, maybe the docs should be updated to reflect this? 

I'm using version 1.0.0.

YoEight:
Fixes #70

~~That PR exposes `Iterable` instead of a `List`, as it used to. I am not totally convinced by this approach but at least, it makes the situation better.~~

~~I'm actually considering rolling back to an observable interface, as the first time we released the Java client, without adding any extra dependency though.~~

It uses a functional (FP) applicative `Fold` abstraction. While very composable, it also allows the user to plug different library, like a reactive framework on top of it, if they feel to.

This implementation allows multiple folds composition, without sacrificing performance. See `FoldTests.java` for example.

We provide smart constructors to help usage. All the code examples have been updated accordingly.


YoEight:
@johanhaleby @daMupfel you are welcomed to jump in!

thefringeninja (approved):

johanhaleby:
It's definitely an improvement with Iterable over List, thank you! But may I ask why you've decided to go with Iterable and not a Stream? Stream has some nice operators (filter, map, and all the rest) that are handy.

YoEight:
@johanhaleby in the meantime, we went beyond that. Now I chose a 'reactive' approach. I implemented a bunch of functional abstractions around it like `fold`.

I didn't use a `Stream` interface because I have no idea how to implement it on top of what the java gRPC is using. The current interface should allow you to return a `Stream` if you know how to do it.

johanhaleby:
A reactive approach would probably be the best approach (although maybe you want both since I guess that most java devs are not familiar with reactive programming). 

You can convert an Iterable to a Stream by doing `StreamSupport.stream(iterable.spliterator(), false);`.

daMupfel:
First of all, great that this is being worked on :).

With reactive approach i assume we mean implementing a Publisher? For java 8 support that would require to load the compat libray. But then you can choose your reactivity framework of choice.
See: https://www.reactive-streams.org/

YoEight:
@johanhaleby 

The Java gRPC library doesn't give an `Iterable`. It gives you something very similar to the `Observer` pattern you found in reactive framework.

@daMupfel 

I don't want to add an external dependency for it but I'm 100% in favour of facilitating their integration from a user perspective.

daMupfel:
I agree that external libraries are usually to be thought carefully about. I to think that the reactive-java way is a bit special since it is part of java 9+. The library is basically backport.

hayley-jean:
This PR is on hold while we review how we want to handle large streams like this in our other clients

chrispoulsen:
Some sort of reactive interface would make sense, I think it translates nicely to how the client-server transport actually happens and seems to align well with the subscription concept that is already present.

I was pointed here after feeling that there was an inconsistency in the API wrt. reading a stream of events (it seems the other clients already have some way of streaming the results). Having to wait for the entire stream to be returned as a List and then immediately convert it back to a stream of some sort, just feels awkward. 

I have not had much experience using the `java.util.concurrent.Flow` interfaces, but I guess they are OK and would go with them, unless there is some compelling reason not to. 

Java 8 compatibility is not a priority for me (considering that the Java17 LTS is almost here, requiring v11 would be fine) - but it could probably be handled by providing the reactive stuff as a separate client extension that requires java 11 (for example) like many other projects have done over the years (it may require some effort to avoid split packages etc. depending on how hard you encapsulate).

Edit: Btw. simply returning a Stream would go along way compared to the current situation.

YoEight:
Done here: https://github.com/EventStore/EventStoreDB-Client-Java/pull/106

PR #84, opened 2021-07-08
Persistent subscriptions to $all

pvanbuijtene:
Adds: Support for subscribing to persistent subscriptions to $all
Fixes: Persistent streams to start from the end by default.

**Breaking:** As defined by the _Persistent Subscriptions RFC_, the initial value for a persistent stream has been changed to start from the `end`.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
@@ -1,0 +1,172 @@
+package com.eventstore.dbclient;
+
+import java.time.Duration;
+
+public abstract class AbstractPersistentSubscriptionSettings<T> {
+    private int checkpointAfterMs;
+    private boolean extraStatistics;
+    private boolean resolveLinkTos;
+    private int historyBufferSize;
+    private int liveBufferSize;
+    private int checkPointUpperBound;
+    private int maxRetryCount;
+    private int maxSubscriberCount;
+    private int messageTimeoutMs;
+    private int checkPointLowerBound;
+    private int readBatchSize;
+    private String consumerStrategyName;
+
+    public AbstractPersistentSubscriptionSettings(int checkpointAfterMs, boolean extraStatistics, boolean resolveLinks,
+                                                  int historyBufferSize, int liveBufferSize, int checkPointUpperBound,
+                                                  int maxRetryCount, int maxSubscriberCount, int messageTimeoutMs,
+                                                  int minCheckpointCount, int readBatchSize, String consumerStrategyName) {
+        this.checkpointAfterMs = checkpointAfterMs;
+        this.extraStatistics = extraStatistics;
+        this.resolveLinkTos = resolveLinks;
+        this.historyBufferSize = historyBufferSize;
+        this.liveBufferSize = liveBufferSize;
+        this.checkPointUpperBound = checkPointUpperBound;
+        this.maxRetryCount = maxRetryCount;
+        this.maxSubscriberCount = maxSubscriberCount;
+        this.messageTimeoutMs = messageTimeoutMs;
+        this.checkPointLowerBound = minCheckpointCount;
+        this.readBatchSize = readBatchSize;
+        this.consumerStrategyName = consumerStrategyName;
+    }
+
+    /**
+     * The amount of time in milliseconds to try to checkpoint after.
+     */
+    public int getCheckpointAfterMs() {
+        return checkpointAfterMs;
+    }
+
+    /**
+     * The amount of time to try to checkpoint after.
+     */
+    public Duration getCheckpointAfter() {
+        return Duration.ofMillis(checkpointAfterMs);
+    }
+
+    /**
+     * Whether to track latency statistics on this subscription.
+     */
+    public boolean isExtraStatistics() {
+        return extraStatistics;
+    }
+
+    /**
+     * @deprecated prefer {@link #shouldResolveLinkTos()}
+     */
+    public boolean isResolveLinks() {
+        return shouldResolveLinkTos();
+    }
+
+    /**
+     * Whether the subscription should resolve linkTo events to their linked events.
+     */
+    public boolean shouldResolveLinkTos() {
+        return resolveLinkTos;
+    }
+
+    /**
+     * The number of events to cache when catching up. Default 500.
+     */
+    public int getHistoryBufferSize() {
+        return historyBufferSize;
+    }
+
+    /**
+     * The size of the buffer (in-memory) listening to live messages as they happen before paging occurs. Default 500.
+     */
+    public int getLiveBufferSize() {
+        return liveBufferSize;
+    }
+
+    /**
+     * @deprecated prefer {@link #getCheckPointUpperBound()}
+     */
+    @Deprecated
+    public int getMaxCheckpointCount() {
+        return this.getCheckPointUpperBound();
+    }
+
+    /**
+     * The maximum number of messages not checkpointed before forcing a checkpoint.
+     */
+    public int getCheckPointUpperBound() {
+        return checkPointUpperBound;
+    }
+
+    /**
+     * The maximum number of retries (due to timeout) before a message is considered to be parked.
+     */
+    public int getMaxRetryCount() {
+        return maxRetryCount;
+    }
+
+    /**
+     * The maximum number of subscribers allowed.
+     */
+    public int getMaxSubscriberCount() {
+        return maxSubscriberCount;
+    }
+
+    /**
+     * The amount of time after which to consider a message as timed out and retried.
+     */
+    public Duration getMessageTimeout() {
+        return Duration.ofMillis(messageTimeoutMs);
+    }
+
+    /**
+     * The amount of time in milliseconds after which to consider a message as timed out and retried.
+     */
+    public int getMessageTimeoutMs() {
+        return messageTimeoutMs;
+    }
+
+    /**
+     * @deprecated prefer {@link #getCheckPointLowerBound()}
+     */
+    @Deprecated
+    public int getMinCheckpointCount() {
+        return this.getCheckPointLowerBound();
+    }
+
+    /**
+     * The minimum number of messages to process before a checkpoint may be written.
+     */
+    public int getCheckPointLowerBound() {
+        return checkPointLowerBound;
+    }
+
+    /**
+     * The number of events read at a time when catching up.
+     */
+    public int getReadBatchSize() {
+        return readBatchSize;
+    }
+
+    /**
+     * @deprecated prefer {@link #getConsumerStrategyName()}
+     */
+    @Deprecated
+    public ConsumerStrategy getStrategy() throws Exception {
+
+        switch (consumerStrategyName){
+            case NamedConsumerStrategy.DISPATCH_TO_SINGLE: return ConsumerStrategy.DispatchToSingle;
+            case NamedConsumerStrategy.ROUND_ROBIN: return ConsumerStrategy.RoundRobin;
+            case NamedConsumerStrategy.PINNED: return ConsumerStrategy.Pinned;
+        }
+
+        throw new Exception("Non-default ConsumerStrategy specified, use getConsumerStrategyName()");
+    }
+
+    /**
+     * The strategy to use for distributing events to client consumers.
+     */
+    public String getConsumerStrategyName() {
+        return consumerStrategyName;
+    }
+}
--- a/db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
@@ -1,0 +1,235 @@
+package com.eventstore.dbclient;
+
+import java.time.Duration;
+
+public class AbstractPersistentSubscriptionSettingsBuilder<T> {
+    protected int checkpointAfterMs;
+    protected boolean extraStatistics;
+    protected boolean resolveLinkTos;
+    protected int historyBufferSize;
+    protected int liveBufferSize;
+    protected int checkPointUpperBound;
+    protected int maxRetryCount;
+    protected int maxSubscriberCount;
+    protected int messageTimeoutMs;
+    protected int checkPointLowerBound;
+    protected int readBatchSize;
+    protected String consumerStrategyName;
+
+    public AbstractPersistentSubscriptionSettingsBuilder() {
+        checkpointAfterMs = 2_000;
+        resolveLinkTos = false;
+        extraStatistics = false;
+        messageTimeoutMs = 30_000;
+        maxRetryCount = 10;
+        checkPointLowerBound = 10;
+        checkPointUpperBound = 1_000;
+        maxSubscriberCount = 0;
+        liveBufferSize = 500;
+        readBatchSize = 20;
+        historyBufferSize = 500;
+        consumerStrategyName = NamedConsumerStrategy.ROUND_ROBIN;
+    }
+
+    public AbstractPersistentSubscriptionSettingsBuilder(AbstractPersistentSubscriptionSettings settings) {
+        checkpointAfterMs = settings.getCheckpointAfterMs();
+        resolveLinkTos = settings.shouldResolveLinkTos();
+        extraStatistics = settings.isExtraStatistics();
+        messageTimeoutMs = settings.getMessageTimeoutMs();
+        maxRetryCount = settings.getMaxRetryCount();
+        checkPointLowerBound = settings.getCheckPointLowerBound();
+        checkPointUpperBound = settings.getCheckPointUpperBound();
+        maxSubscriberCount = settings.getMaxSubscriberCount();
+        liveBufferSize = settings.getLiveBufferSize();
+        readBatchSize = settings.getReadBatchSize();
+        historyBufferSize = settings.getHistoryBufferSize();
+        consumerStrategyName = settings.getConsumerStrategyName();
+    }
+
+    /**
+     * @deprecated prefer {@link #resolveLinkTos()}
+     */
+    @Deprecated
+    public T enableLinkResolution() {
+        return resolveLinks(true);
+    }
+
+    /**
+     * @deprecated prefer {@link #notResolveLinkTos()}
+     */
+    @Deprecated
+    public T disableLinkResolution() {
+        return resolveLinks(false);
+    }
+
+    /**
+     * @deprecated prefer {@link #resolveLinkTos(boolean)}
+     */
+    @Deprecated
+    public T resolveLinks(boolean value) {
+        return (T) resolveLinkTos(value);
+    }
+
+    /**
+     * Whether the subscription should resolve linkTo events to their linked events. Default: false.
+     */
+    public T resolveLinkTos(boolean value) {
+        this.resolveLinkTos = value;
+        return (T) this;
+    }
+
+    /**
+     * Resolve linkTo events to their linked events.
+     */
+    public T resolveLinkTos() {
+        return this.resolveLinkTos(true);
+    }
+
+    /**
+     * Don't resolve linkTo events to their linked events.
+     */
+    public T notResolveLinkTos() {
+        return this.resolveLinkTos(false);
+    }
+
+    /**
+     * Enable tracking of latency statistics on this subscription.
+     */
+    public T enableExtraStatistics() {
+        return extraStatistics(true);
+    }
+
+    /**
+     * Disable tracking of latency statistics on this subscription.
+     */
+    public T disableExtraStatistics() {
+        return extraStatistics(false);
+    }
+
+    /**
+     * Whether to track latency statistics on this subscription. Default: false.
+     */
+    public T extraStatistics(boolean value) {
+        this.extraStatistics = value;
+        return (T) this;
+    }
+
+    /**
+     * The amount of time to try to checkpoint after. Default: 2 seconds.
+     */
+    public T checkpointAfter(Duration value) {
+        this.checkpointAfterInMs((int)value.toMillis());
+        return (T) this;
+    }
+
+    /**
+     * The amount of time in milliseconds to try to checkpoint after. Default: 2 seconds.
+     */
+    public T checkpointAfterInMs(int value) {
+        this.checkpointAfterMs = value;
+        return (T) this;
+    }
+
+    /**
+     * The number of events to cache when catching up. Default: 500.
+     */
+    public T historyBufferSize(int value) {
+        this.historyBufferSize = value;
+        return (T) this;
+    }
+
+    /**
+     * The size of the buffer (in-memory) listening to live messages as they happen before paging occurs. Default: 500.
+     */
+    public T liveBufferSize(int value) {
+        this.liveBufferSize = value;
+        return (T) this;
+    }
+
+    /**
+     * @deprecated prefer {@link #checkPointUpperBound(int)}
+     */
+    @Deprecated
+    public T maxCheckpointCount(int value) {
+        return checkPointUpperBound(value);
+    }
+
+    /**
+     * The maximum number of messages not checkpointed before forcing a checkpoint. Default: 1000.
+     */
+    public T checkPointUpperBound(int value) {
+        this.checkPointUpperBound = value;
+        return (T) this;
+    }
+
+    /**
+     * @deprecated prefer {@link #checkPointLowerBound(int)}
+     */
+    @Deprecated
+    public T minCheckpointCount(int value) {
+        return this.checkPointLowerBound(checkPointLowerBound);
+    }
+
+    /**
+     * The minimum number of messages to process before a checkpoint may be written. Default: 10.
+     */
+    public T checkPointLowerBound(int value) {
+        this.checkPointLowerBound = value;
+        return (T) this;
+    }
+
+    /**
+     * The maximum number of subscribers allowed. Default: 0 (Unbounded).
+     */
+    public T maxSubscriberCount(int value) {
+        this.maxSubscriberCount = value;
+        return (T) this;
+    }
+
+    /**
+     * The maximum number of retries (due to timeout) before a message is considered to be parked. Default: 10.
+     */
+    public T maxRetryCount(int value) {
+        this.maxRetryCount = value;
+        return (T) this;
+    }
+
+    /**
+     * The amount of time after which to consider a message as timed out and retried. Default: 30 seconds.
+     */
+    public T messageTimeout(Duration value) {
+        return this.messageTimeoutInMs((int)value.toMillis());
+    }
+
+    /**
+     * The amount of time in milliseconds after which to consider a message as timed out and retried. Default: 30 seconds.
+     */
+    public T messageTimeoutInMs(int value) {
+        this.messageTimeoutMs = value;
+        return (T) this;
+    }
+
+    /**
+     * The number of events read at a time when catching up. Default: 20.
+     */
+    public T readBatchSize(int value) {
+        this.readBatchSize = value;
+        return (T) this;
+    }
+
+    /**
+     * The strategy to use for distributing events to client consumers.
+     */
+    public T consumerStrategy(ConsumerStrategy strategy) {
+        this.consumerStrategyName = NamedConsumerStrategy.from(strategy);
+        return (T) this;
+    }
+
+    /**
+     * The strategy to use for distributing events to client consumers.
+     */
+    public T namedConsumerStrategy(String value) {
+        this.consumerStrategyName = value;
+        return (T) this;
+    }
+}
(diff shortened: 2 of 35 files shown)


YoEight (approved):
LGTM. Don't forget to squash your commits before merging

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        this.maxSubscriberCount = maxSubscriberCount;
+        this.messageTimeoutMs = messageTimeoutMs;
+        this.minCheckpointCount = minCheckpointCount;
+        this.readBatchSize = readBatchSize;
+        this.fromStart = fromStart;
+        this.fromEnd = fromEnd;
+        this.strategy = strategy;
+    }
+
+    public int getCheckpointAfterMs() {
```
oskardudycz, line 40: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R51) 
```suggestion
    public int getCheckpointAfter() {
```
Similar comment below: https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR20

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        return checkpointAfterMs;
+    }
+
+    public boolean isExtraStatistics() {
+        return extraStatistics;
+    }
+
+    public boolean isResolveLinks() {
```
oskardudycz, line 61: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R43). And it's not aligned with other options. See more in e.g. [OptionsWithResolveLinkTosBase](https://github.com/EventStore/EventStoreDB-Client-Java/blob/trunk/db-client-java/src/main/java/com/eventstore/dbclient/OptionsWithResolveLinkTosBase.java#L19)
```suggestion
    public boolean resolveLinkTos() {
```
The same comment applies e.g. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR61.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public int getHistoryBufferSize() {
+        return historyBufferSize;
+    }
+
+    public int getLiveBufferSize() {
+        return liveBufferSize;
+    }
+
+    public int getMaxCheckpointCount() {
```
oskardudycz, line 90: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R43).
```suggestion
    public int getCheckPointUpperBound() {
```
The same comment applies e.g. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR28.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public int getMaxRetryCount() {
+        return maxRetryCount;
+    }
+
+    public int getMaxSubscriberCount() {
+        return maxSubscriberCount;
+    }
+
+    public int getMessageTimeoutMs() {
```
oskardudycz, line 125: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R46).
```suggestion
    public int getMessageTimeout() {
```
The same comment applies e.g. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR25.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        return messageTimeoutMs;
+    }
+
+    public int getMinCheckpointCount() {
```
oskardudycz, line 133: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R52).
```suggestion
    public int getCheckPointLowerBound() {
```
The same comment applies e.g. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR27.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        return minCheckpointCount;
+    }
+
+    public int getReadBatchSize() {
+        return readBatchSize;
+    }
+
+    public boolean getFromStart() {
```
oskardudycz: **CONCERN:** I think it would be more straightforward to introduce enum with the explicit values and have a single getter. Accordingly, for the setters in the builders, we can have a setter with enum plus the syntactic sugar as we have right now. 
Also, shouldn't we allow to set the specific position? 
See more in the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R44).

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        return fromStart;
+    }
+
+    public boolean getFromEnd() {
+        return fromEnd;
+    }
+
+    public ConsumerStrategy getStrategy() {
```
oskardudycz: **FIX:** Setting does not match the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R55).
```suggestion
    public ConsumerStrategy getConsumerStrategyName() {
```
The same comment applies e.g. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-b38c322fb2ef3a75dc3289f2babf28c7359937f0530d73c8f3a3734ac18b4dceR33.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+    protected int minCheckpointCount;
+    protected int readBatchSize;
+    protected ConsumerStrategy strategy;
+
+    public AbstractPersistentSubscriptionSettingsBuilder() {
+        checkpointAfterMs = 2_000;
+        resolveLinks = false;
+        extraStatistics = false;
+        fromStart = false;
+        fromEnd = false;
```
oskardudycz: **FIX:** Based on the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R44) the default starting point should be the end.
```suggestion
        fromEnd = true;
```

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+        minCheckpointCount = settings.getMinCheckpointCount();
+        maxCheckpointCount = settings.getMaxCheckpointCount();
+        maxSubscriberCount = settings.getMaxSubscriberCount();
+        liveBufferSize = settings.getLiveBufferSize();
+        readBatchSize = settings.getReadBatchSize();
+        historyBufferSize = settings.getHistoryBufferSize();
+        strategy = settings.getStrategy();
+    }
+
+    public T enableLinkResolution() {
```
oskardudycz, line 53: **FIX:** ResolveLinkTos naming is not aligned with other options. See more e.g. in [OptionsWithResolveLinkTosBase](https://github.com/EventStore/EventStoreDB-Client-Java/blob/trunk/db-client-java/src/main/java/com/eventstore/dbclient/OptionsWithResolveLinkTosBase.java#L19).

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
```diff
@@ -0,0 +1,141 @@
+package com.eventstore.dbclient;
+
+public class AbstractPersistentSubscriptionSettingsBuilder<T> {
```
oskardudycz, line 5: See related comments in [AbstractPersistentSubscriptionSettings](https://github.com/EventStore/EventStoreDB-Client-Java/pull/84/files#diff-3f22e62cf5819afff2ad131e5e7058c61029a3bc78c5490c78d1f057b8ebd9dbR3).

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public T fromEnd() {
+        this.fromEnd = true;
+        this.fromStart = false;
+        return (T) this;
+    }
+
+
+    public T historyBufferSize(int value) {
```
oskardudycz, line 136: **SUGGESTION:** I think that typically in Java, such setters should start with the prefix `set` or the name should be an imperative form.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractUpdatePersistentSubscription.java
```diff
@@ -0,0 +1,73 @@
...
+                    .setMaxCheckpointCount(settings.getMaxCheckpointCount())
+                    .setMessageTimeoutMs(settings.getMessageTimeoutMs())
+                    .setMaxSubscriberCount(settings.getMaxSubscriberCount())
+                    .setMaxRetryCount(settings.getMaxRetryCount())
+                    .setLiveBufferSize(settings.getLiveBufferSize())
+                    .setHistoryBufferSize(settings.getHistoryBufferSize())
+                    .setExtraStatistics(settings.isExtraStatistics())
+                    .setCheckpointAfterMs(settings.getCheckpointAfterMs());
+
+            switch (settings.getStrategy()) {
```
oskardudycz: **CONCERN:** Aren't we missing `PinnedByCorrelation` here? See more in the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R71).

oskardudycz:
@pvanbuijtene, @hayley-jean I'm not sure if that's out of the scope of this PR, but I don't see the implementation of:
- the `AutoAck` [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R62),
- `ReplayParkedMessages` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
- `RestartPersistentSubscriptionSubsystem` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
- `GetInfo` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R26),
- `ListAll` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R29),
- `ListForStream`method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R30).
- `PinnedByCorrelation` strategy mentioned here: https://github.com/EventStore/EventStoreDB-Client-Java/pull/84#discussion_r678078940.

@hayley-jean, could you wage what should go here and what's not? 

I guess that probably `AutoAck` should go in this one. The rest can go in the dedicated, follow-up PRs.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
hayley-jean: We could also have an overload that takes a duration without the `Ms` in the name

db-client-java/src/test/java/com/eventstore/dbclient/DeletePersistentSubscriptionTests.java
```diff
@@ -15,4 +11,13 @@ public void testDeletePersistentSub() throws Throwable {
...
     }
+
+    @Test
+    public void testDeletePersistentSubToAll() throws Throwable {
+        client.createToAll("aGroupUpd")
+                .get();
+
+        client.deleteToAll("aGroupUpd")
+                .get();
+    }
```
oskardudycz, line 22: **CONCERN:** Shouldn't we have some assertion that verifies if the operations succeeded? We can have the false-positive test results if delete silently fails. Similar comment to other added/changed tests.

db-client-java/src/test/java/com/eventstore/dbclient/SubscribePersistentSubcription.java
```diff
@@ -82,4 +82,59 @@ public void onCancelled(PersistentSubscription subscription) {
 
         Assert.assertEquals(6, result.get().intValue());
     }
+
```
oskardudycz, line 85: **SUGGESTION:** Can we add test scenarios for the negative cases, e.g.:
- what happens if subscribing to a not available database,
- nodes were disconnected during the persistent subscription liveness,
- error while handling the persistent subscription logic,
- etc.

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBPersistentSubscriptionsClient.java
```diff
@@ -11,20 +11,52 @@ public static EventStoreDBPersistentSubscriptionsClient create(EventStoreDBClien
...
     public CompletableFuture create(String stream, String group) {
-        return this.create(stream, group, CreatePersistentSubscriptionOptions.get());
+        return this.createToStream(stream, group);
     }
 
+    public CompletableFuture createToAll(String group) {
+        return this.createToAll(group, CreatePersistentSubscriptionToAllOptions.get());
+    }
+
+    public CompletableFuture createToStream(String stream, String group) {
```
oskardudycz: **CONCERN:** Based on the [RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R18), this method is redundant, as we can just use `create`. The same applies to other `ToStream` management methods.

oskardudycz (changes requested):
See the comments above.

oskardudycz:
Also, documentation snippets are missing. See the .NET snippets for persistent subscriptions: https://github.com/EventStore/EventStore-Client-Dotnet/blob/master/samples/persistent-subscriptions/Program.cs.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
pvanbuijtene: @oskardudycz `getConsumerStrategyName` suggests a string is being returned, I know it's the name from the RFC but would `getConsumerStrategy` not be better since an enum is returned ?

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: Agree, but in this case it's a builder which normally has a more fluent syntax?

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: This was already in-place, deprecated it and linked to new `resolveLinkTos()`.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
pvanbuijtene: Added an overload which returns the `Duration`.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
pvanbuijtene: I refactored the `fromStart` and `fromEnd`, the implementations can be found in the specific classes. The position can be set by the builder and with the constructor.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: The default for streams is already defined to begin from the start, do we want to change this?
I changed it for the `all` stream subscription.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: This was already in-place, deprecated it and linked to new method.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: Agree, but this is a builder class which normally has a more fluent naming?

db-client-java/src/test/java/com/eventstore/dbclient/DeletePersistentSubscriptionTests.java
pvanbuijtene: This is how the current tests are written, currently the `List`/`Get` operations aren't implemented yet for persistent subscriptions. I would suggest to implement this in separate PR and modify/add the tests.

db-client-java/src/test/java/com/eventstore/dbclient/SubscribePersistentSubcription.java
pvanbuijtene: Agree, see previous comment. https://github.com/EventStore/EventStoreDB-Client-Java/pull/84#discussion_r678663827

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBPersistentSubscriptionsClient.java
pvanbuijtene: Removed `toStream` classes and methods.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractUpdatePersistentSubscription.java
pvanbuijtene: Think this is not implemented yet, can't find this on the server ?

pvanbuijtene:
> @pvanbuijtene, @hayley-jean I'm not sure if that's out of the scope of this PR, but I don't see the implementation of:
> 
> * the `AutoAck` [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R62),
> * `ReplayParkedMessages` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
> * `RestartPersistentSubscriptionSubsystem` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
> * `GetInfo` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R26),
> * `ListAll` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R29),
> * `ListForStream`method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R30).
> * `PinnedByCorrelation` strategy mentioned here: [Persistent subscriptions to $all #84 (comment)](https://github.com/EventStore/EventStoreDB-Client-Java/pull/84#discussion_r678078940).
> 
> @hayley-jean, could you wage what should go here and what's not?
> 
> I guess that probably `AutoAck` should go in this one. The rest can go in the dedicated, follow-up PRs.

`AutoAck` is also not implemented for normal persistent subscriptions, suggest to do this in separate PR.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
hayley-jean: I think if it's an enum being returned `getConsumerStrategy` makes the most sense

db-client-java/src/main/java/com/eventstore/dbclient/AbstractUpdatePersistentSubscription.java
hayley-jean: PinnedByCorrelation isn't available on the server proto yet, so should come in a later PR

hayley-jean:
> @pvanbuijtene, @hayley-jean I'm not sure if that's out of the scope of this PR, but I don't see the implementation of:
> 
> * the `AutoAck` [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R62),
> * `ReplayParkedMessages` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
> * `RestartPersistentSubscriptionSubsystem` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R25),
> * `GetInfo` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R26),
> * `ListAll` method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R29),
> * `ListForStream`method [mentioned in the RFC](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R30).
> * `PinnedByCorrelation` strategy mentioned here: [Persistent subscriptions to $all #84 (comment)](https://github.com/EventStore/EventStoreDB-Client-Java/pull/84#discussion_r678078940).
> 
> @hayley-jean, could you wage what should go here and what's not?
> 
> I guess that probably `AutoAck` should go in this one. The rest can go in the dedicated, follow-up PRs.

The missing operations still need to be added to the server proto before we can implement them (see [here](https://github.com/EventStore/home/pull/550/files#diff-ad3a37ec8a5416ddcd905109248d13dba6a72a352ce7ee41b9d3d004aae1ff95R75))

AutoAck could go in this PR or in a follow-up one that also implements it for the other persistent subscriptions.

YoEight (approved):

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettings.java
oskardudycz: I think that having both methods is confusing. The general convention is to not include `Ms` in the name as all our timeouts should be set in milliseconds.

oskardudycz:
@hayley-jean, I think that we should add an issue to remove all deprecated fields in the next major release.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
oskardudycz: I think that we should be aligned with the RFC also for the starting stream position. It will be a breaking change.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
oskardudycz: It is not the hill I'd die for, but IMHO, having set in this case would be more explicit in this case. But it's not a blocker for me to keep it as it is.

db-client-java/src/main/java/com/eventstore/dbclient/AbstractUpdatePersistentSubscription.java
oskardudycz: It might be worth creating an issue for that to not forget about it.

db-client-java/src/test/java/com/eventstore/dbclient/DeletePersistentSubscriptionTests.java
oskardudycz: I'm good with adding them in the separate PR, but it'd be good to create an issue for that to not forget about it.

oskardudycz (approved):
@pvanbuijtene, thanks for the fixes. Approving 👍 

It'd be good to clarify https://github.com/EventStore/EventStoreDB-Client-Java/pull/84#discussion_r682376481 before merging.

db-client-java/src/test/java/com/eventstore/dbclient/DeletePersistentSubscriptionTests.java
pvanbuijtene: Done https://github.com/EventStore/EventStoreDB-Client-Java/issues/87

db-client-java/src/main/java/com/eventstore/dbclient/AbstractUpdatePersistentSubscription.java
pvanbuijtene: Done https://github.com/EventStore/EventStoreDB-Client-Java/issues/88

db-client-java/src/main/java/com/eventstore/dbclient/AbstractPersistentSubscriptionSettingsBuilder.java
pvanbuijtene: Done.

pvanbuijtene:
**Issues created:**
Remove deprecated methods and classes https://github.com/EventStore/EventStoreDB-Client-Java/issues/86
Improve code coverage for persistent subscriptions https://github.com/EventStore/EventStoreDB-Client-Java/issues/87
Support PinnedByCorrelation strategy for persistent subscriptions https://github.com/EventStore/EventStoreDB-Client-Java/issues/88

PR #98, opened 2021-09-02
Add equals() and hashCode() to SpecificExpectedRevision

dpasek-senacor:
Add `equals()` and `hashCode()` to SpecificExpectedRevision class to allow quality checks.

These methods are missing and therefore it is not possible to properly check the equality of two `ExpectedRevision` instances.
Adding proper `equals()` implementation is helpful for implementing logic and tests regarding the expected revisions.


YoEight (changes requested):
Thanks for your contribution!

It looks good but I think you can also implement `equals` and `hashCode` for all `ExpectedRevision` sub-classes too.

dpasek-senacor:
@YoEight Thanks for your quick review.

The reason why I skipped the implementation of `equals()` and `hashCode()` on the other subclasses was them being singletons. The constructors are package private and therefore a client is not able to create new instances. A client can only use the singleton instances providing instance equality. 

But you are right: for the internal code of the client it would be helpful to be on the safe side and have alle concrete subsclasses to implement equals() / hashCode().

YoEight (approved):

oskardudycz (approved):

YoEight:
Great work!

PR #99, opened 2021-09-03
Unify error handling behavior for subscriptions.

dpasek-senacor:
The error handling behavior between persistent and non-persistent subscriptions is inconsistent and partly erroneous:
- Persistent subscriptions do not handle the "requires leader" case.
- Non persistent subscriptions call onError() twice in the case of `NotLeaderException`

This pull requests unifies the behavior and avoids duplicate calls to the error listener.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/AbstractRegularSubscription.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/AbstractRegularSubscription.java
@@ -96,23 +96,23 @@ public abstract class AbstractRegularSubscription {
 
                 @Override
                 public void onError(Throwable throwable) {
-                    if (throwable instanceof StatusRuntimeException) {
-                        StatusRuntimeException e = (StatusRuntimeException) throwable;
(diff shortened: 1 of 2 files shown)


YoEight (approved):
Please stop making my life easier :smile:

YoEight:
Nice catch by the way!

dpasek-senacor:
@YoEight 
>Please stop making my life easier

😂  

Maybe just as an explanation: we are currently preparing the migration of our EventStoreDB integration from the old (unofficial) ESJC (see: https://github.com/msemys/esjc no longer actively developed as TCP protocol is deprecated) to your new gRPC Java client. Our integration has quite an extensive test coverage and therefore it is quite easy for us to find issues in the EventStoreDB client.
As we want to proceed with the migration it is also easier for me to provide the patches myself instead of writing issue reports or feature requests and hoping things will be done by other people. 😉

oskardudycz (approved):
Changes look good to me 👍  Thank you for the contribution!

@YoEight, we should probably improve our test coverage in this area.

PR #101, opened 2021-09-08
Improve node selection algorithm

### LINKED ISSUE #83: nodePreference seems to always connect to the same node
I've noticed the following behaviour with nodePreference:

`nodePreference=follower` - seems to always connect to the same follower node instead of randomly choosing one among them (stopping and starting the client makes it connect to the same node every time)
`nodePreference=random` - seems to always connect to the same node (among all nodes that are currently alive)
`nodePreference=leader` - works fine since there's a single leader

connection string:
```
esdb://admin:changeit@127.0.0.1:2113,127.0.0.2:2113,127.0.0.3:2113?tls=true&tlsVerifyCert=false&nodePreference=NODEPREFERENCE
```

I've tested with version 1.0.0 and `trunk`

dpasek-senacor:
This fixes https://github.com/EventStore/EventStoreDB-Client-Java/issues/83

Currently the node selection strategy always produces stable result always prefering a single node of the specified category. Always the same follower or read-only replica is selected based on the `ClusterInfo` retrieved from the seed node.

Even the `NodePreference.RANDOM` currently will always result in the same node to be returned to the client if the `Member` in the `ClusterInfo` will not change.

This PR properly performs random selection of nodes in each category. It also extracts the algorithm of the node selection into dedicated class for better testability.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/NodeSelector.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/NodeSelector.java
@@ -1,0 +1,72 @@
+package com.eventstore.dbclient;
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Random;
+import java.util.Set;
+
+class NodeSelector {
+
+    private static final Random random = new Random();
+    private static final Set<ClusterInfo.MemberState> invalidStates;
+    private static final Map<NodePreference,Comparator<ClusterInfo.Member>> memberComparators;
+
+    static {
+        invalidStates = new HashSet<ClusterInfo.MemberState>() {{
+            add(ClusterInfo.MemberState.MANAGER);
+            add(ClusterInfo.MemberState.SHUTTING_DOWN);
+            add(ClusterInfo.MemberState.SHUT_DOWN);
+            add(ClusterInfo.MemberState.UNKNOWN);
+            add(ClusterInfo.MemberState.INITIALIZING);
+            add(ClusterInfo.MemberState.RESIGNING_LEADER);
+            add(ClusterInfo.MemberState.PRE_LEADER);
+            add(ClusterInfo.MemberState.PRE_REPLICA);
+            add(ClusterInfo.MemberState.PRE_READ_ONLY_REPLICA);
+            add(ClusterInfo.MemberState.CLONE);
+            add(ClusterInfo.MemberState.DISCOVER_LEADER);
+        }};
+
+        memberComparators= new HashMap<NodePreference, Comparator<ClusterInfo.Member>>()  {{
+            put(NodePreference.LEADER, new MemberComparator(ClusterInfo.MemberState.LEADER));
+            put(NodePreference.FOLLOWER, new MemberComparator(ClusterInfo.MemberState.FOLLOWER));
+            put(NodePreference.READ_ONLY_REPLICA, new MemberComparator(ClusterInfo.MemberState.READ_ONLY_REPLICA));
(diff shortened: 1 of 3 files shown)


YoEight (approved):
Solid work!

dpasek-senacor:
@YoEight 
I have added some more test cases on the `NodeSelector`

Theres is a miniscule chance that tests break if rolling the dice 100 times on node selection is not enough. 😉

db-client-java/src/main/java/com/eventstore/dbclient/NodeSelector.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Set;
+
+class NodeSelector {
+
+    private static final Random random = new Random();
+    private static final Set<ClusterInfo.MemberState> invalidStates;
+    private static final Map<NodePreference,Comparator<ClusterInfo.Member>> memberComparators;
+
+    static {
+        invalidStates = new HashSet<ClusterInfo.MemberState>() {{
```
oskardudycz, line 18: **CONCERN:** I think that: `PreReadOnlyReplica` and `ReadOnlyLeaderless` might be valid state for the `ReadOnlyReplica` preference.

oskardudycz (approved):
Neat work @dpasek-senacor!

@YoEight @hayley-jean, we'll also need to tune it a bit to be aligned with the Reconnection RFC, e.g.
- use allowed list instead of invalid,
- allow `PreReadOnlyReplica`, `ReadOnlyLeaderless` for `ReadOnlyReplica` note preference,
- If no nodes match the NodePreference, then move any Leader nodes to the top of the list.
- Trying to perform an operation that requires a leader when you have NodePreference that specifically can't connect to a leader should result in a client error.

Still, this is a great contribution. We can do it as a follow-up 👍

db-client-java/src/main/java/com/eventstore/dbclient/NodeSelector.java
dpasek-senacor: @oskardudycz 
I agree, but it is a subsequent issue as this implementation is based on status quo in trunk:
https://github.com/EventStore/EventStoreDB-Client-Java/blob/trunk/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClusterClient.java#L22
https://github.com/EventStore/EventStoreDB-Client-Java/blob/trunk/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClusterClient.java#L94

oskardudycz: Yup, I think that we could expand it in the follow-up PR together with other alignments I mentioned inhttps://github.com/EventStore/EventStoreDB-Client-Java/pull/101#pullrequestreview-754754366. This is more than good enough for now :)

ahjohannessen:
>we'll also need to tune it a bit to be aligned with the Reconnection RFC

@oskardudycz Where can I find this RFC? :)

oskardudycz:
@ahjohannessen so far is internal, but I hope that we'll be able to show it publicly soon. If not through RFC, then maybe at least through docs 🙂

ahjohannessen:
@oskardudycz ok, would be nice to have such that unofficial clients have some guide lines to follow :)

oskardudycz:
@ahjohannessen, we want to be more open, not only in regards to the results and ask in advance for suggestions. To do that, we have to sort the RFC process internally (almost there 😉 ) then find a sustainable way to make that public. I'm sure that this will happen soon (no exact timeframe tho).

Regarding specs for unofficial clients, that's also on our TODO list. We won't be able to maintain all the languages, so for sure, we're taking that into account to work closely with the community 👍

PR #104, opened 2021-10-20
Implement simple reactive interface for reads.

YoEight:


db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
```diff
@@ -46,76 +47,97 @@ public static EventStoreDBClient create(EventStoreDBClientSettings settings) {
...
+                return observer.getStreamMetadata();
             }
 
             throw new RuntimeException(e);
         });
     }
 
-    public CompletableFuture<ReadResult> readAll() {
-        return this.readAll(Long.MAX_VALUE, ReadAllOptions.get());
+    public <A> CompletableFuture<A> readAll(ReadObserver<A> observer) {
```
oskardudycz, line 121: **SUGGESTION:** I think we should also add an overload without the observer and set the `new EventCollectorReadObserver()` for that case. It will be more accessible for the people using classically to get all events from the short stream.

oskardudycz:
@dpasek-senacor, could you also add a test verifying if the changes helped and it's possible to enumerate through the long stream (e.g. a few thousand events)?

YoEight:
Done here: https://github.com/EventStore/EventStoreDB-Client-Java/pull/106

PR #106, opened 2021-10-26
Implementation of reactive streaming API for read operations.

dpasek-senacor:
This would a suggestion on implementing the streaming of read operations based on the reactive streams API supporting full backpressure and out-of-the-box compatibility with reative frameworks like RxJava, Project Reactor and JDK 9 Flow.

This solution consists of the following parts:
- Adaption of the EventStoreDBClient-API to return a (cold) `Publisher` of events on read operations. This is implemented by the `AbstractRead` (and its sub classes implementing the `Publisher` contract using the internal `ReadSubscription` class for fulfilling the reactive contract.
- Introduction of reactive streams TCK as test suite for testing the read operations and the reactive contract. 

The major building block is definetly the `ReadSubscription` class which acts as the adapter between the gRPC `StreamObserver`-API and the `Publisher` / `Subscriber` contract. This implementation is currently based on thread synchronisation between the gRPC reader thread used for triggering the `StreamObserver` and the `Subscribers` consumer thread which can be any thread including the main thread.

### DIFF
--- a/db-client-java/build.gradle
+++ b/db-client-java/build.gradle
@@ -30,6 +30,8 @@ configurations {
 }
 
 dependencies {
+    api "org.reactivestreams:reactive-streams:${reactiveStreamsApiVersion}"
+
     implementation "javax.annotation:javax.annotation-api:${annotationApiVersion}"
     implementation "javax.validation:validation-api:${validationApiVersion}"
 
@@ -43,8 +45,9 @@ dependencies {
 
     testImplementation "junit:junit:${junitVersion}"
     testImplementation "org.slf4j:slf4j-nop:${slf4jNopVersion}"
+    testImplementation "io.reactivex.rxjava3:rxjava:3.1.2"
+    testImplementation "org.reactivestreams:reactive-streams-tck:${reactiveStreamsApiVersion}"
     testImplementation "org.testcontainers:testcontainers:${testcontainersVersion}"
-
     testImplementation platform("com.fasterxml.jackson:jackson-bom:${jacksonVersion}")
     testImplementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 }
--- a/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
@@ -1,8 +1,12 @@
 package com.eventstore.dbclient;
 
+import org.reactivestreams.Publisher;
+
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 public class EventStoreDBClient extends EventStoreDBClientBase {
@@ -59,13 +63,29 @@ public class EventStoreDBClient extends EventStoreDBClientBase {
     }
 
     public CompletableFuture<ReadResult> readStream(String streamName, long maxCount, ReadStreamOptions options) {
+        return readEventsFromPublisher(this.readStreamReactive(streamName, maxCount, options));
+    }
+
+    public Publisher<ResolvedEvent> readStreamReactive(String streamName) {
+        return this.readStreamReactive(streamName, Long.MAX_VALUE, ReadStreamOptions.get());
+    }
+
+    public Publisher<ResolvedEvent> readStreamReactive(String streamName, long maxCount) {
+        return this.readStreamReactive(streamName, maxCount, ReadStreamOptions.get());
+    }
+
+    public Publisher<ResolvedEvent> readStreamReactive(String streamName, ReadStreamOptions options) {
+        return this.readStreamReactive(streamName, Long.MAX_VALUE, options);
+    }
+
+    public Publisher<ResolvedEvent> readStreamReactive(String streamName, long maxCount, ReadStreamOptions options) {
         if (options == null)
             options = ReadStreamOptions.get();
 
         if (!options.hasUserCredentials())
             options.authenticated(this.credentials);
 
-        return new ReadStream(this.client, streamName, maxCount, options).execute();
+        return new ReadStream(this.client, streamName, maxCount, options);
     }
 
     public CompletableFuture<StreamMetadata> getStreamMetadata(String streamName) {
@@ -109,13 +129,29 @@ public class EventStoreDBClient extends EventStoreDBClientBase {
     }
 
     public CompletableFuture<ReadResult> readAll(long maxCount, ReadAllOptions options) {
+        return readEventsFromPublisher(this.readAllReactive(maxCount, options));
+    }
+
+    public Publisher<ResolvedEvent> readAllReactive() {
+        return this.readAllReactive(Long.MAX_VALUE, ReadAllOptions.get());
+    }
+
+    public Publisher<ResolvedEvent> readAllReactive(long maxCount) {
+        return this.readAllReactive(maxCount, ReadAllOptions.get());
+    }
+
+    public Publisher<ResolvedEvent> readAllReactive(ReadAllOptions options) {
+        return this.readAllReactive(Long.MAX_VALUE, options);
+    }
+
+    public Publisher<ResolvedEvent> readAllReactive(long maxCount, ReadAllOptions options) {
         if (options == null)
             options = ReadAllOptions.get();
 
         if (!options.hasUserCredentials())
             options.authenticated(this.credentials);
 
-        return new ReadAll(this.client, maxCount, options).execute();
+        return new ReadAll(this.client, maxCount, options);
     }
 
     public CompletableFuture<Subscription> subscribeToStream(String streamName, SubscriptionListener listener) {
@@ -158,5 +194,27 @@ public class EventStoreDBClient extends EventStoreDBClientBase {
             options.authenticated(this.credentials);
 
         return new DeleteStream(this.client, streamName, options).execute();
+    }
+
+    private static CompletableFuture<ReadResult> readEventsFromPublisher(Publisher<ResolvedEvent> eventPublisher) {
+        CompletableFuture<ReadResult> future = new CompletableFuture<>();
+        List<ResolvedEvent> events = new LinkedList<>();
+        eventPublisher.subscribe(new ReadSubscriber() {
+            @Override
+            public void onEvent(ResolvedEvent resolvedEvent) {
+                events.add(resolvedEvent);
+            }
+
+            @Override
+            public void onError(Throwable t) {
+                future.completeExceptionally(t);
+            }
+
+            @Override
+            public void onComplete() {
+                future.complete(new ReadResult(events));
+            }
+        });
+        return future;
     }
 }
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -13,6 +13,8 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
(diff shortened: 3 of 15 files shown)


db-client-java/build.gradle
```diff
@@ -30,6 +30,8 @@ configurations {
 }
 
 dependencies {
+    api "org.reactivestreams:reactive-streams:${reactiveStreamsApiVersion}"
```
dpasek-senacor, line 33: This is also a good practice in gradle to mark the dependencies used in APIs of a library as `api` dependencies.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -38,7 +40,15 @@ protected GrpcClient(EventStoreDBClientSettings settings, SslContext sslContext)
 
     protected void startConnectionLoop() {
         pushMsg(new CreateChannel(this.currentChannelId));
-        CompletableFuture.runAsync(this::messageLoop);
+        CompletableFuture.runAsync(this::messageLoop, createConnectionLoopExecutor());
```
dpasek-senacor, line 43: See other pull requests. Was necessary for TCK tests:
https://github.com/EventStore/EventStoreDB-Client-Java/pull/105

db-client-java/src/main/java/com/eventstore/dbclient/ReadSubscriber.java
```diff
@@ -0,0 +1,26 @@
+package com.eventstore.dbclient;
+
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+public abstract class ReadSubscriber implements Subscriber<ResolvedEvent> {
```
dpasek-senacor: Simple base class as a subscriber _not_ supporting backpressure. Currently only internally used. Shouldn't probably part of the public API of this client.

db-client-java/src/main/java/com/eventstore/dbclient/ReadSubscription.java
```diff
@@ -0,0 +1,67 @@
...
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+class ReadSubscription implements Subscription {
```
dpasek-senacor, line 15: This is the major class providing the backpressure handling between the gRPC client and the `Subscriber`.

db-client-java/src/test/java/com/eventstore/dbclient/ReactiveStreamsPublisherVerificationTCK.java
```diff
@@ -0,0 +1,276 @@
...
+import org.reactivestreams.tck.PublisherVerification;
+import org.reactivestreams.tck.TestEnvironment;
+import org.testng.SkipException;
+
+import java.lang.reflect.Method;
+
+import static org.junit.Assert.assertNotNull;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public abstract class ReactiveStreamsPublisherVerificationTCK<T> extends PublisherVerification<T> {
```
dpasek-senacor, line 17: JUnit wrapper for the reactive stream API TCK which is written in TestNg.

db-client-java/src/main/java/com/eventstore/dbclient/RecordedEvent.java
```diff
@@ -112,6 +114,19 @@ public String getContentType() {
         return contentType;
     }
 
+    @Override
+    public boolean equals(Object o) {
```
dpasek-senacor, line 118: Necessary for TCK testing: tests assures that multiple subscribers receive the same events if subscribing to the same publisher.

db-client-java/src/test/java/com/eventstore/dbclient/ReadAllPublisherVerification.java
```diff
@@ -0,0 +1,46 @@
+package com.eventstore.dbclient;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.tck.TestEnvironment;
+import testcontainers.module.EventStoreTestDBContainer;
+
+public class ReadAllPublisherVerification extends ReactiveStreamsPublisherVerificationTCK<ResolvedEvent> {
```
dpasek-senacor, line 10: Implementation of the TCK for the ReadAll operations.

db-client-java/src/test/java/com/eventstore/dbclient/ReadAllTests.java
```diff
@@ -20,10 +25,12 @@ public void testReadAllEventsForwardFromZeroPosition() throws ExecutionException
                 .fromStart()
                 .notResolveLinkTos();
 
-        ReadResult result = client.readAll(10, options)
-                .get();
+        List<ResolvedEvent> result = Flowable.fromPublisher(client.readAll(10, options))
```
dpasek-senacor: For simple testing and asserting reactive stream we now can use the RxJava testing framework.

YoEight (approved):

db-client-java/src/test/java/com/eventstore/dbclient/samples/quick_start/QuickStart.java
```diff
@@ -43,10 +42,22 @@ public static void Run() throws ParseError, ExecutionException, InterruptedExcep
                 .forwards()
                 .fromStart();
 
-        ReadResult result = client.readStream("some-stream", 10, options)
-                .get();
+        client.readStream("some-stream", 10, options).subscribe(new ReadSubscriber() {
```
oskardudycz: **SUGGESTION:** This sample intent is to show how to read a set of events from the stream, so it should be more like the one above: https://github.com/EventStore/EventStoreDB-Client-Java/pull/106/files#diff-c53c867ba6cee9fc84ea5583415293244d6371732798e517b74de2fe8523e770R96

Similar suggestion for the samples below.

oskardudycz:
I reviewed @dpasek-senacor PR. It's excellent work. Thank you for the enhanced comments guiding about the changes.

I'll start with a disclaimer: I'm haven't worked on a Java project for a few years, so my perspective may be skewed. I have a general concern about the PR. I'm all for enabling the reactive approach and giving the option to enumerate through the long streams. I know that the RX approach is popular in the Java land, but I'm not sure how much that's a standard. I wonder if we could provide an API or a bit of syntactic sugar, making the default scenario so reading the whole stream more accessible? Based on my experience, most people want to read the entire stream and do left-fold to aggregate the state. I know that we could use EventCollectorReadSubscriber, but the user always needs to add a few additional lines. I'm not sure how much that's doable, but maybe we could come up with the API that's not an RX-first and enable accessibility on both scenarios?

Thoughts?

dpasek-senacor:
Hi @oskardudycz 

Thanks for the review and your thoughts about the approach.

>I'm all for enabling the reactive approach and giving the option to enumerate through the long streams. I know that the RX approach is popular in the Java land, but I'm not sure how much that's a standard. 

The approach using the reactive stream API should be the most standard way currently possible. It is not an RX approach, i.e. not tied to any special implementation of reactive streams but uses the API that allows interoperability with the market standards: RxJava and Project Reactor. It is also API compatible to the new JDK 9 `Flow` API which is the new official standard for consuming streams with backpressure. As discussed with @YoEight internally we would propose not to use `Flow`-API directly as this would break JDK8 support. This would be a bit to early as many projects are still tied to JDK 8. In 2 years it would be easy to migrate to JDK9 `Flow`-API by just replacing imports without changing any code.
Last but not least this API is the closest to the API used in other languages, e.g. .net (using `IAsyncEnumerable<ResolvedEvent>` as return type) or JavaScript (using `StreamingRead<AllStreamResolvedEvent>`)

>I wonder if we could provide an API or a bit of syntactic sugar, making the default scenario so reading the whole stream more accessible? Based on my experience, most people want to read the entire stream and do left-fold to aggregate the state. I know that we could use EventCollectorReadSubscriber, but the user always needs to add a few additional lines. I'm not sure how much that's doable, but maybe we could come up with the API that's not an RX-first and enable accessibility on both scenarios?

I think that the given approach provides us the flexibility and feature set and give us the options to make the live easier for certain use cases:
- You have full backpressure support. Backpressure is necessary on reading and processing long streams. Getting backpressure right is not a simple tasks. It is always beneficial to let the clinet/driver handle the dirty details of it.
-  You can easily use the power of RxJava or Project Reactor (filtering, mapping, grouping and parallelism) if needed. 
- If we think that reading the events into a list is a standard operations we should add it to the client as a method, i.e. create an API on EventStoreDBClient, that clearly separates "read" and "stream" operations. This would be very simple to add using the foundation we have now. The old Memsys ESJC had this approach and it was very helpful to let the developer decide  if streaming and reading is the better approach.

So yes we should consider adding methods which provide this "read into list" support as currently available, but use the reactive foundation behind it. This would also be beneficial in regard of keeping the next release of the driver backward compatible to the 1.0 release. I'm a huge fan of semantic versioning and following this approach the current change would require to release it as "breaking change", i.e. a major version. Keeping the old method signatures and adding the new ones for streaming would prevent this.

oskardudycz:
@dpasek-senacor, I agree that for the long streams proposed, API is a great way to deal with them (especially regarding efficient memory usage and backpressure you mentioned). 

My main take was that typically, you should try to avoid such cases, where you have really long streams. Of course, sometimes it's unavoidable, so it should be possible to do that effectively. We should try to provide the change non-breaking, so adding streams methods keeps the old one. After that, we could try to gather feedback from the community if they're okay with removing the support for the old ones.

dpasek-senacor:
@YoEight @oskardudycz 
I have now updated the access to streams via reactive streams to be an additional option in the API next to the existing classic read methods. Internally it uses the reactive infrastracture. 
I have also reverted the samples to reference the previous implementation.

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
```diff
@@ -59,13 +63,29 @@ public static EventStoreDBClient create(EventStoreDBClientSettings settings) {
     }
 
     public CompletableFuture<ReadResult> readStream(String streamName, long maxCount, ReadStreamOptions options) {
+        return readEventsFromPublisher(this.streamEvents(streamName, maxCount, options));
+    }
+
+    public Publisher<ResolvedEvent> streamEvents(String streamName) {
```
oskardudycz: **SUGGESTION:** Thinking out loud, `streamEvents` may be confused with the general stream concept. Maybe we could name those methods like `readStreamAsync`.

@dpasek-senacor, @YoEight, @hayley-jean, @jageall thoughts?

oskardudycz (commented):
Neat work @dpasek-senacor! Looks good to me. I'll approve once we have confirmation on the final naming for the new methods

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
dpasek-senacor: `readStreamAsync()` would be fine for me.

dpasek-senacor: Being precise... the other `readStream()` methods are also async (using CompletableFuture) 😉 

Internally we discussed `readStreamReactive` or `readStreamReactively` but it felt pretty clunky, `readStreamAsync` is definetely better.

oskardudycz: Yup, I know, it's tricky, for `subscribeToStream` would also work well, but it's also clashing with the subscription feature... 😉

 `readStreamReactive` is also okay for me. We'll need to discuss how to make this API more self-explanatory and consistent in the long term. But that's another story. 😉

timothycoleman: `readStreamReactive` seems sensible to me

dpasek-senacor: @YoEight @hayley-jean @jageall 
Any more thoughts on the naming of the reactive methods?
Current two contenders: 
- `readStreamAsync` / `readAllAsync` 
- `readStreamReactive` / `readAllReactive`

YoEight: we are choosing the `Reactive` suffix!

YoEight (approved):

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
dpasek-senacor: Done!

oskardudycz (approved):
Great work, @dpasek-senacor!

oskardudycz (approved):

YoEight (approved):

PR #107, opened 2021-10-27
Force reconnect when receiving aborted status code.

YoEight:
Force reconnect when receiving aborted status code.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -93,7 +93,7 @@ public abstract class GrpcClient {
                 if (error instanceof StatusRuntimeException) {
                     StatusRuntimeException ex = (StatusRuntimeException) error;
 
-                    if (ex.getStatus().getCode().equals(Status.Code.UNAVAILABLE)) {
+                    if (ex.getStatus().getCode().equals(Status.Code.UNAVAILABLE) || ex.getStatus().getCode().equals(Status.Code.ABORTED)) {
                         self.pushMsg(new CreateChannel(id));
                     }
                 }


db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -93,7 +93,7 @@ protected void pushMsg(Msg msg) {
                 if (error instanceof StatusRuntimeException) {
                     StatusRuntimeException ex = (StatusRuntimeException) error;
 
-                    if (ex.getStatus().getCode().equals(Status.Code.UNAVAILABLE)) {
+                    if (ex.getStatus().getCode().equals(Status.Code.UNAVAILABLE) || ex.getStatus().getCode().equals(Status.Code.ABORTED)) {
```
oskardudycz, line 96: **QUESTION:** Is it possible to write a test simulating this scenario?

YoEight: A consistent one? Nope. But it happens from time-to-time on the Go client cluster tests.

oskardudycz: Couldn't it be simulated by killing the node and bringing it back? I know that @George-Payne did some tests like that in NodeJS.

George-Payne: It tests `UNAVAILABLE`, not sure how you would force an `ABORTED`:
https://github.com/EventStore/EventStore-Client-NodeJS/blob/master/src/__test__/connection/reconnect.test.ts

oskardudycz (commented):
Approving it to unblock the client, but tests around that should be tackled as a separate issue.

oskardudycz (approved):

PR #122, opened 2021-12-15
Expose filter options when creating a persistent subscription.

YoEight:


db-client-java/src/main/java/com/eventstore/dbclient/SubscriptionFilter.java
```diff
@@ -86,5 +87,56 @@ void addToWireReadReq(StreamsOuterClass.ReadReq.Options.Builder builder) {
 
         builder.setFilter(optsB.build());
     }
+
+    void addToWirePersistentReadReq(Persistent.CreateReq.AllOptions.Builder builder) {
```
oskardudycz: **SUGGESTION:** It might be worth aligning at least part of the logic with the regular subscription processing, as it's almost the same.

oskardudycz:
There are no tests checking the filtering feature.

oskardudycz (changes requested):
I could approve it if it was the single blocker, but I think that we should add at least basic test coverage for the feature.

oskardudycz (approved):

pvanbuijtene (approved):

PR #138, opened 2022-02-28
Implement server feature detection.

YoEight:


db-client-java/build.gradle
```diff
@@ -43,13 +43,26 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.11.1'
     implementation "org.slf4j:slf4j-api:1.7.30"
 
-    testImplementation "junit:junit:${junitVersion}"
-    testImplementation "org.slf4j:slf4j-nop:${slf4jNopVersion}"
+    testImplementation "org.junit.jupiter:junit-jupiter-api:5.8.2"
+    testImplementation "org.junit.jupiter:junit-jupiter-params:5.8.2"
+    testImplementation "org.junit.jupiter:junit-jupiter-engine:5.8.2"
+    testImplementation "org.junit.platform:junit-platform-suite:1.8.2"
+//    testImplementation "org.slf4j:slf4j-nop:${slf4jNopVersion}"
```
pvanbuijtene: Looks like this can be removed?

db-client-java/src/main/java/com/eventstore/dbclient/ServerInfo.java
```diff
@@ -0,0 +1,15 @@
+package com.eventstore.dbclient;
+
+public class ServerInfo {
+    private final ServerVersion version;
+    private final int features;
+
+    public ServerInfo(ServerVersion version, int features) {
+        this.version = version;
```
pvanbuijtene, line 8: It looks like the value hold by`ServerVersion` is never used in the code, maybe it can be entirely removed?

YoEight: It might have some usages later on but not in this PR. I rather keep it.

db-client-java/build.gradle
YoEight: Done

pvanbuijtene (approved):

George-Payne (approved):

PR #142, opened 2022-03-25
Apply Deadline RFC

YoEight:


db-client-java/src/main/java/com/eventstore/dbclient/OptionsBase.java
```diff
@@ -5,10 +5,15 @@
 class OptionsBase<T> {
     protected Timeouts timeouts;
```
oskardudycz: **CONCERN:** Shouldn't we remove timeouts and use deadlines now? I think that they're not applicable after RFC. Also, I think that we should update default values.

YoEight: Done

db-client-java/src/main/java/com/eventstore/dbclient/GossipClient.java
```diff
@@ -25,7 +26,7 @@ public GossipClient(ManagedChannel channel) {
     }
 
     public void shutdown() throws InterruptedException {
-        _channel.shutdown().awaitTermination(Timeouts.DEFAULT.shutdownTimeout, Timeouts.DEFAULT.shutdownTimeoutUnit);
+        _channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
oskardudycz, line 29: **CONCERN:** Shouldn't it be 10 seconds? See RFC: https://github.com/EventStore/architecture-and-planning/blob/main/rfcs/032%20-%20grpc-clients-deadlines.md#default-timeouts

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -276,7 +280,7 @@ private void messageLoop() {
     private void closeConnection() {
         if (this.channel != null) {
             try {
-                this.channel.shutdown().awaitTermination(Timeouts.DEFAULT.shutdownTimeout, Timeouts.DEFAULT.shutdownTimeoutUnit);
+                this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
oskardudycz, line 283: The same concern as above: https://github.com/EventStore/EventStoreDB-Client-Java/pull/142/files#r853930699.

oskardudycz:
I think that it's worth adding some tests to make sure that timeouts are correctly applied on the gRPC channels/connection, etc.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
YoEight: I would argue that deadlines are for operations only and nothing in the RFC mentions anything for connection management. The Java gRPC lib is the only one with an async close function. I set five because it was the previous default value for shutdowns.

db-client-java/src/main/java/com/eventstore/dbclient/GossipClient.java
YoEight: I would argue that deadlines are for operations only and nothing in the RFC mentions anything for connection management. The Java gRPC lib is the only one with an async close function. I set five because it was the previous default value for shutdowns.

YoEight:
I fail to see the value behind it, to be honest. For some reason, the Java gRPC lib decided to have an async method for closing a channel. I also feel I would test the internal gRPC client implementation too.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
oskardudycz: The RFC, in my opinion, states the clear deadlines as 10 seconds. @hayley-jean, thoughts?

oskardudycz:
> I fail to see the value behind it, to be honest. For some reason, the Java gRPC lib decided to have an async method for closing a channel. I also feel I would test the internal gRPC client implementation too.

I think that we at least need to find repeatable proof in tests that those settings are applied as in the written RFC. I know that it may be non-trivial to have on the integration tests, but maybe there is a way to test it accordingly a least as the removed timeout builder tests.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
hayley-jean: The deadline only applies to the operations though, this looks like it's a timeout for closing the gRPC connection which isn't covered by any of the RFCs.

If the previous default was 5 seconds, then I think it would be best not to change it

YoEight:
Added a couple of extra deadline tests

oskardudycz (approved):
Nice one @YoEight, thank you for adding tests for deadlines 👍

hayley-jean (approved):

PR #144, opened 2022-04-27
Expose current stream positions and retry count when consuming a subscriptions

YoEight:
This PR exposes:
* the first and last event number of a stream when reading (regular read and subscriptions)
* retry count when consuming a persistent subscription.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
@@ -1,6 +1,7 @@
 package com.eventstore.dbclient;
 
 import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
 
 import java.util.Arrays;
 import java.util.HashMap;
@@ -8,6 +9,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
 
 public class EventStoreDBClient extends EventStoreDBClientBase {
     private EventStoreDBClient(EventStoreDBClientSettings settings) {
@@ -63,22 +65,19 @@ public class EventStoreDBClient extends EventStoreDBClientBase {
     }
 
     public CompletableFuture<ReadResult> readStream(String streamName, long maxCount, ReadStreamOptions options) {
-        return readEventsFromPublisher(this.readStreamReactive(streamName, maxCount, options));
+        Publisher<ReadMessage> publisher = readStreamReactive(streamName, maxCount, options);
+        return readEventsFromPublisher(publisher);
     }
 
-    public Publisher<ResolvedEvent> readStreamReactive(String streamName) {
+    public Publisher<ReadMessage> readStreamReactive(String streamName) {
         return this.readStreamReactive(streamName, Long.MAX_VALUE, ReadStreamOptions.get());
     }
 
-    public Publisher<ResolvedEvent> readStreamReactive(String streamName, long maxCount) {
-        return this.readStreamReactive(streamName, maxCount, ReadStreamOptions.get());
-    }
-
-    public Publisher<ResolvedEvent> readStreamReactive(String streamName, ReadStreamOptions options) {
+    public Publisher<ReadMessage> readStreamReactive(String streamName, ReadStreamOptions options) {
         return this.readStreamReactive(streamName, Long.MAX_VALUE, options);
     }
 
-    public Publisher<ResolvedEvent> readStreamReactive(String streamName, long maxCount, ReadStreamOptions options) {
+    public Publisher<ReadMessage> readStreamReactive(String streamName, long maxCount, ReadStreamOptions options) {
         if (options == null)
             options = ReadStreamOptions.get();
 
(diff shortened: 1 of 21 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
```diff
@@ -78,7 +80,7 @@ public Publisher<ResolvedEvent> readStreamReactive(String streamName, ReadStream
         return this.readStreamReactive(streamName, Long.MAX_VALUE, options);
     }
 
-    public Publisher<ResolvedEvent> readStreamReactive(String streamName, long maxCount, ReadStreamOptions options) {
+    public Publisher<ResolvedEventWithContext> readStreamReactiveWithContext(String streamName, long maxCount, ReadStreamOptions options) {
```
oskardudycz: **QUESTION:** Do we intend to merge this method with `readStreamReactive` in the next major release? The same question applies to other methods.

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
```diff
@@ -1,12 +1,15 @@
 package com.eventstore.dbclient;
 
 public abstract class PersistentSubscriptionListener {
-    public void onEvent(PersistentSubscription subscription, ResolvedEvent event) {
+    public void onEvent(PersistentSubscription subscription, int retryCount, Position position, ResolvedEvent event) {
```
oskardudycz: **QUESTION:** Why is the position not a part of `ResolvedEvent` but a separate parameter?

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
```diff
@@ -1,12 +1,15 @@
...
     }
 
     public void onError(PersistentSubscription subscription, Throwable throwable) {
     }
 
     public void onCancelled(PersistentSubscription subscription) {
     }
+
+    public void onConfirmation(PersistentSubscription subscription) {
```
oskardudycz, line 13: **QUESTION:** Is this callback invoked after ACK?

db-client-java/src/main/java/com/eventstore/dbclient/ReadSubscription.java
```diff
@@ -43,13 +43,13 @@ public void onError(Throwable error) {
         subscriber.onError(error);
     }
 
-    public void onNext(ResolvedEvent event) {
+    public void onNext(long firstStreamPosition, long lastStreamPosition, Position lastAllStreamPosition, ResolvedEvent event) {
```
oskardudycz: **SUGGESTION:** Could we wrap those positions into object? It should be easier to provide non-breaking changes for that in the future.

db-client-java/src/main/java/com/eventstore/dbclient/SubscriptionListener.java
```diff
@@ -1,12 +1,16 @@
 package com.eventstore.dbclient;
 
 public abstract class SubscriptionListener {
-    public void onEvent(Subscription subscription, ResolvedEvent event) {
+    public void onEvent(Subscription subscription, SubscriptionContext context, ResolvedEvent event) {
```
oskardudycz: **CONCERN:** This seems to be a breaking change. Are we okay with that? If we'd like to keep that non-breaking, the potential option is to provide a context wrapper for the subscription parameter. It could derive and/or expose the same public properties as Subscription plus the additional position information. Then we wouldn't need to provide the additional parameter. Technically it'd be still binary-breaking change, but that less severe than regular breaking change. Thoughts?

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
YoEight: Nope, it's called when the server confirms the subscription has started. There is a specific gRPC message for that.

db-client-java/src/main/java/com/eventstore/dbclient/ReadSubscription.java
YoEight: This class is not publicly exposed.

db-client-java/src/main/java/com/eventstore/dbclient/SubscriptionListener.java
YoEight: This is a breaking change indeed. There are already some breaking changes that got merged a few days ago so I won't worried about that too much.

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
YoEight: Done

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
YoEight: Done.

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
```diff
@@ -145,7 +170,7 @@ public Publisher<ResolvedEvent> readAllReactive(ReadAllOptions options) {
         return this.readAllReactive(Long.MAX_VALUE, options);
     }
 
-    public Publisher<ResolvedEvent> readAllReactive(long maxCount, ReadAllOptions options) {
+    public Publisher<ResolvedEventWithContext> readAllReactiveWithContext(long maxCount, ReadAllOptions options) {
```
oskardudycz: **CONCERN:** @YoEight, as you derived `ResolvedEventWithContext` from `ResolvedEvent`, then this method seems redundant. I think that it's okay to just have `readAllReactive` that returns `ResolvedEventWithContext`. Am I missing some important point here? :)

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
oskardudycz: Maybe we could name it `onStarted`? Do you know if/how other clients are naming it?

oskardudycz (commented):
@YoEight, @hayley-jean, I'm not against such changes; they look fine. However, I think that we should make sure that this is the intended handling also in other clients. It is already different from .NET gRPC client. I'm not sure how other clients are dealing with it.

db-client-java/src/main/java/com/eventstore/dbclient/PersistentSubscriptionListener.java
hayley-jean: The only client that I can find that exposes this is the NodeJS one, which uses `confirmation`, but the other clients use `confirmation` internally. I'd be inclined to keep this named the same as the message coming over gRPC for now

db-client-java/src/main/java/com/eventstore/dbclient/EventStoreDBClient.java
YoEight: Done.

hayley-jean (approved):

thefringeninja (approved):

Issue #151, opened 2022-06-06
Retrieval of server features breaks connection retry


dpasek-senacor:
*Affected version*: 3.0

*Steps to reproduce*: Use a single node configuration to an event store server which is down or starting up.

*Expected behavior*: The client performs a retry of connecting to the server until a connection is established or the number of retries configured for the client is reached.
*Observed behavior*: The client does not perform a retry on creating a connection to the server. On first attempt the client fails and performs a shutdown.

The introduction of the retrieval of the supported server features is the root cause of the issue for the single node client. A single node client implements the abstract `doConnect()` method simply by creating a Netty channel to the configured server host. This operation will always succeed, as this operation - in my understanding of Netty - will not actually create an connection to the server but just creates the channel infrastructure. The acual HTTP connection is created during the first RPC command execution.
Since `doConnect()`always succeeds the next operation will be retrieving the server features, which will fail with `StatusRuntimeException UNAVAILABLE` as a connection to the server cannot be established. But instead of retrying the operation the client is shutdown and will be unusable.

See the following code of the `discover()`-Method in the `GrpcClient` class with comments:
```Java
        for (; ; ) {
            logger.debug("Start connection attempt ({}/{})", attempts, settings.getMaxDiscoverAttempts());
            if (doConnect()) { // Will always succeed even if server is not started in single node setup
                try {
                    serverInfo = ServerFeatures.getSupportedFeatures(settings, channel); // Will fail, when server is not started yet.
                } catch (Exception e) {
                    logger.error("An exception happened when fetching server supported features", e);
                    return false;  // Causes the client to shutdown and do not perform a retry on the connection
                }
                currentChannelId = UUID.randomUUID();

                logger.info("Connection created successfully");

                return true;
            } else {
                ++attempts;
                if (attempts > settings.getMaxDiscoverAttempts()) {
                    logger.error("Maximum discovery attempt count reached: {}", settings.getMaxDiscoverAttempts());
                    return false;
                }

                logger.warn("Unable to find a node. Retrying... ({}/{})", attempts, settings.getMaxDiscoverAttempts());
                sleep(settings.getDiscoveryInterval());
            }
        }
```

I will prepare an PR for fixing this behavior, as it breaks integration tests, where the server is started asynchronously and may not be ready, when the client instance is created.

dpasek-senacor:
Works fine now. Issue has been fixed.

PR #152, opened 2022-06-06
Issue 151: Fix connection attempts

dpasek-senacor:
See: https://github.com/EventStore/EventStoreDB-Client-Java/issues/151

Fixes the retry behavior on connecting to EventStoreDB server, in case the server features could not be retrieved, as server is currently unavailble.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -141,42 +141,51 @@ abstract class GrpcClient {
         if (candidate.isPresent()) {
             this.endpoint = candidate.get();
             this.channel = createChannel(this.endpoint);
+
             try {
-                this.serverInfo = ServerFeatures.getSupportedFeatures(this.settings, this.channel);
+                if (loadServerFeatures()) {
+                    this.currentChannelId = UUID.randomUUID();
+                    return true;
+                }
             } catch (Exception e) {
-                logger.error("An exception happened when fetching server supported features", e);
-                return false;
+                logger.error("A fatal exception happened when fetching server supported features", e);
             }
-
-            this.currentChannelId = UUID.randomUUID();
-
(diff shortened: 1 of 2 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -141,27 +141,17 @@ private boolean discover(UUID previousId, Optional<Endpoint> candidate) {
...
         for (; ; ) {
             logger.debug("Start connection attempt ({}/{})", attempts, settings.getMaxDiscoverAttempts());
-            if (doConnect()) {
-                try {
-                    serverInfo = ServerFeatures.getSupportedFeatures(settings, channel);
-                } catch (Exception e) {
-                    logger.error("An exception happened when fetching server supported features", e);
-                    return false;
-                }
+            if (doConnect() && loadServerFeatures()) {
```
dpasek-senacor: Both steps must be successful for initializing the client, otherwise the retry loop will be used.

YoEight (changes requested):
Thanks, @dpasek-senacor for your contribution.

I think a better approach would be to decide whether the server features retrieval process is retriable or not. Whatever exception you got when your ESDB server is not ready/up is a valid case for retrying or moving to the next candidate.

dpasek-senacor:
> I think a better approach would be to decide whether the server features retrieval process is retriable or not. Whatever exception you got when your ESDB server is not ready/up is a valid case for retrying or moving to the next candidate.

Hi @YoEight ! I have adapted the PR to take the type of exception into account on fetching the server features. This will now throw a `RetryableException` if a retry should be performed otherwise the client is shutdown as before.

I don't know if this is what you meant, but now only a `TimeoutException` or Status `UNAVAIBLE` will trigger the retry.

YoEight (approved):

ylorph (approved):

Issue #155, opened 2022-06-13
Improvement: Shutdown of client should not be blocked by pending connection attempts


dpasek-senacor:
Currently the shutdown of a `EventStoreDBClient` is blocked until all connection attempts have been completed. 

It would be better if the shutdown of the client will abort the running connection attempts.

The problem is located in the `GrpcClient` class in the `dicover` method:
```Java
    private boolean discover(UUID previousId, Optional<Endpoint> candidate) {
        long attempts = 1;
...   
        for (; ; ) {
            logger.debug("Start connection attempt ({}/{})", attempts, settings.getMaxDiscoverAttempts());
            if (doConnect()) {
                try {
                    serverInfo = ServerFeatures.getSupportedFeatures(settings, channel);
                } catch (Exception e) {
                    logger.error("An exception happened when fetching server supported features", e);
                    return false;
                }
                currentChannelId = UUID.randomUUID();

                logger.info("Connection created successfully");

                return true;
            } else {
                ++attempts;
                if (attempts > settings.getMaxDiscoverAttempts()) {
                    logger.error("Maximum discovery attempt count reached: {}", settings.getMaxDiscoverAttempts());
                    return false;
                }

                logger.warn("Unable to find a node. Retrying... ({}/{})", attempts, settings.getMaxDiscoverAttempts());
                sleep(settings.getDiscoveryInterval());
            }
        }
    }
```
The current implementation will loop all connecton attempts before allowing the client to process other messages like the `Shutdown` message. Depending of the number of attempts und the retry interval this might block a client shutdown for many seconds. With the current defaults an connection attempt to a server which does not respond/is not availble would block for more than 10s: 3 times (500ms Retry-Interval + 3000ms for timeout on reading the server features) = 10500ms until `Shutdown` message will be processed.

PR #156, opened 2022-06-14
Improve Shutdown behavior

dpasek-senacor:
Reconnection attempts are now handled as separate messages allowing shutdown during connection attempts.

This is related to https://github.com/EventStore/EventStoreDB-Client-Java/issues/155

The idea behind this implementation is that the reconnection attempts are performed as multiple `CreateChannel` messages instead of a closed `for`-loop which cannot be aborted by a shutdown message.
This new implementation will perform only a single connection attempt per `CreateChannel` message. If this attempt fails a new message will be send to the message queue. The number of attempts is ontained as state inside the messages.
The shutdown message now can be interleaved between the `CreateChannel` messages. Work items will handled as before: if no channel exists as connection failed these items will be sent back to the end of the queue, which might also mean to be handled after a `Shutdown` messages.

*Example*
Client sends the following messages:
- Create Client (= initial `CreateChannel` message)
- readAll()
- Shutdown

*Behavior before:*
- Client tries to connect to server with the configured number of retry attempts (may take mutiple seconds due to longer timeout on implicit server features command).
- Reconnection attempts fail, Client automatically shuts down
- readAll command is ignored during drain of pending messages
- Shutdown command is ignored is ignored during drain of pending messages

*New behavior*
- Client tries to connect to server for the first time. Connection attempt fails. 
- A new `CreateChannel` message is send to the queue (behind the shutdown message)
- The readAll() command is requeued behind the second `CreateChannel` message as no channel exists
- Shutdown command is executed.
- Second `CreateChannel` message is ignored during drain of pending messages
- readAll command is ignored during drain of pending messages

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -131,10 +131,8 @@ abstract class GrpcClient {
         return runWithArgs(args -> CompletableFuture.completedFuture(args.endpoint));
     }
 
-    private boolean discover(UUID previousId, Optional<Endpoint> candidate) {
-        long attempts = 1;
-
-        // It means we already created a new channel and it was old request.
+    private boolean discover(UUID previousId, long attempts, Optional<Endpoint> candidate) {
+        // It means we already created a new channel, and it was old request.
         if (!currentChannelId.equals(previousId))
             return true;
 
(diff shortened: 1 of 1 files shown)


YoEight (commented):
While I understand the issue that PR tries to solve, it's also an edge-case scenario. I'm afraid with that specific change, applications that are under heavy load might suffer from unbound connection retry time as each new attempt is pushed at the bottom of the queue, every time.

PR #158, opened 2022-06-25
Set `RequiresLeader` header to `true` when node preference is to Leader.

### LINKED ISSUE #157: RequiresLeader seems to be set to False by default
This should set to True by default to be in line with the other clients.

YoEight:
Fixes #157


hayley-jean (approved):

ylorph:
https://github.com/EventStore/architecture-and-planning/blob/main/rfcs/574-connectivity-in-grpc-clients.md#connection-settings

ylorph (approved):

PR #160, opened 2022-07-06
added kotlin support for RecordedEvents to be hydrated by data classes

### LINKED ISSUE #159: Improve Kotlin support
Usage of Kotlin data classes for domain objects is very desirable as it provides immutability, etc.

Currently data classes cannot be used because the jackson deserialization process requires an empty default constructor.

The [jackson-module-kotlin](https://github.com/FasterXML/jackson-module-kotlin) library provides just the hooks needed in order to get this working. It should just be a matter of adding the kotlin module if this library is detected on the classpath.

[This is an example](https://github.com/spring-projects/spring-framework/blob/3dc84c2d92c0fb64e23780f840c563cf4a1e483f/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java#L837-L848) of how Spring Framework does it.

nhajratw:
This fixes #159 -- i don't think it takes care of all the kotlin cases, but does take care of the ones i've run into so far.

Also, i notice there are a lot of different JsonMapper creations all over the code. Probably should consolidate that into a factory of some sort.


hayley-jean:
Hi @nhajratw thank you for the contribution

Unfortunately we have decided previously that we will be removing (de)serialization on behalf of the client going forward. You can read more about it in this comment: https://github.com/EventStore/EventStoreDB-Client-Java/pull/147#issuecomment-1125964595

In light of this, I'm going to close the PR as we won't be able to accept it

PR #162, opened 2022-07-08
Issue 161 Force shutdown after grace period

dpasek-senacor:
Fixed: Performs a forced shutdown of gRPC ManagedChannel if normal shutdown does not terminate the `ManagedChannel`.

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -290,7 +290,10 @@ abstract class GrpcClient {
     private void closeConnection() {
         if (this.channel != null) {
             try {
-                this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+                boolean terminated = this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+                if (!terminated) {
+                    this.channel.shutdownNow().awaitTermination(5, TimeUnit.SECONDS);
+                }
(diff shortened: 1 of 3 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -290,7 +290,10 @@ private void messageLoop() {
     private void closeConnection() {
         if (this.channel != null) {
             try {
-                this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+                boolean terminated = this.channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
thefringeninja, line 293: Code looks good, but I wonder if waiting up to 10 seconds is what we want here.

dpasek-senacor: Hi @thefringeninja! Yes, 10secs would be too much, but it should not happen in the real world. We give 5 seconds of grace period for completing existing GRPC commands, then we force the shutdown, which will occur immediately and should only take a few milliseconds as GRPC library will directly kill all operations.

We could definitely reduce the second `awaitTermination()` time out in the code for the force shutdown (e.g. 1-2 secs), but in reality it will "feel" the same, as we will never reach the timeout.

thefringeninja (approved):

pvanbuijtene (approved):

pvanbuijtene:
@dpasek-senacor thanks for your contribution 👍

PR #165, opened 2022-07-26
Add getServerVersion() api method and update dataset log positions / tests

shaan1337:
Added: getServerVersion() api method
Updated: datasets with real prepare/commit log positions
Updated: read stream tests to cater for log position changes on the server
Fixed: bug in version parsing

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -32,6 +32,7 @@ abstract class GrpcClient {
     protected Exception lastException;
     protected UUID currentChannelId;
     protected Optional<ServerInfo> serverInfo = Optional.empty();
+    private volatile boolean serverInfoLoaded = false;
 
     protected volatile boolean shutdown = false;
 
@@ -182,11 +183,19 @@ abstract class GrpcClient {
     private boolean loadServerFeatures() {
         try {
             serverInfo = ServerFeatures.getSupportedFeatures(settings, channel);
+            serverInfoLoaded = true;
             return true;
         } catch (ServerFeatures.RetryableException e) {
             logger.warn("An exception happened when fetching server supported features. Retrying connection attempt.", e);
             return false;
         }
+    }
+
+    public Optional<ServerVersion> getServerVersion() {
+        if (!serverInfoLoaded)
+            throw new RuntimeException("server info not loaded");
+
+        return serverInfo.map(ServerInfo::getServerVersion);
     }
 
     protected void sleep(long millis) {
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
@@ -44,7 +44,7 @@ class ServerFeatures {
         stub.getSupportedMethods(Shared.Empty.getDefaultInstance(), convertSingleResponse(result, resp -> {
             int major = 0, minor = 0, patch = 0;
 
-            String[] splits = resp.getEventStoreServerVersion().split(".");
+            String[] splits = resp.getEventStoreServerVersion().split("\\.");
             for (int idx = 0; idx < splits.length; idx++) {
                 if (idx > 2) {
                     break;
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
@@ -21,4 +21,29 @@ class ServerVersion {
     public int getPatch() {
         return patch;
     }
+
+    public boolean isGreaterThan(int major, int minor, int patch) {
+        return !isLessThan(major, minor, patch) &&
+                !equals(major, minor, patch);
+    }
+
+    public boolean equals(int major, int minor, int patch) {
+        return this.major == major
+                && this.minor == minor
+                && this.patch == patch;
+    }
+
+    public boolean isLessThan(int major, int minor, int patch) {
+        int cmp;
+        if ((cmp = Integer.compare(this.major, major)) != 0)
+            return cmp < 0;
+
+        if ((cmp = Integer.compare(this.minor, minor)) != 0)
+            return cmp < 0;
+
+        if ((cmp = Integer.compare(this.patch, patch)) != 0)
+            return cmp < 0;
+
+        return false;
+    }
 }
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ServerInfo.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ServerInfo.java
@@ -12,4 +12,5 @@ class ServerInfo {
     public boolean supportFeature(int feature) {
         return (features & feature) != 0;
     }
+    public ServerVersion getServerVersion() { return version; }
 }
--- a/db-client-java/src/test/java/com/eventstore/dbclient/ReadStreamReactiveTests.java
+++ b/db-client-java/src/test/java/com/eventstore/dbclient/ReadStreamReactiveTests.java
@@ -6,6 +6,7 @@ import org.junit.jupiter.api.Test;
 import testcontainers.module.ESDBTests;
 
 import java.util.List;
+import java.util.Optional;
 
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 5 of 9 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
```diff
@@ -44,7 +44,7 @@ private static CompletableFuture<ServerInfo> getSupportedFeaturesInternal(Server
         stub.getSupportedMethods(Shared.Empty.getDefaultInstance(), convertSingleResponse(result, resp -> {
             int major = 0, minor = 0, patch = 0;
 
-            String[] splits = resp.getEventStoreServerVersion().split(".");
+            String[] splits = resp.getEventStoreServerVersion().split("\\.");
```
YoEight, line 47: It's currently working, why does it need to be changed?

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -182,13 +183,26 @@ private boolean discover(UUID previousId, Optional<Endpoint> candidate) {
...
             serverInfo = ServerFeatures.getSupportedFeatures(settings, channel);
+            serverInfoLoaded = true;
             return true;
         } catch (ServerFeatures.RetryableException e) {
             logger.warn("An exception happened when fetching server supported features. Retrying connection attempt.", e);
             return false;
         }
     }
 
+    public ServerVersion getServerVersion() throws RuntimeException {
```
YoEight: You don't have to declare that a function would throw a `RuntimeException`.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -182,13 +183,26 @@ private boolean discover(UUID previousId, Optional<Endpoint> candidate) {
...
+        if (!serverInfoLoaded) {
+            throw new RuntimeException("server info not loaded");
```
YoEight, line 196: Considering that logic was not present before, why are we throwing an exception in this case?

db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
```diff
@@ -4,6 +4,9 @@ class ServerVersion {
     private final int minor;
     private final int patch;
 
+    // old server with no support for server features api
+    public static ServerVersion Old = new ServerVersion(0, 0, 0);
```
YoEight: I would rather keep the current logic where if there is no server version available, we don't return one with bogus values in it.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -32,6 +32,7 @@ abstract class GrpcClient {
     protected Exception lastException;
     protected UUID currentChannelId;
     protected Optional<ServerInfo> serverInfo = Optional.empty();
+    private volatile boolean serverInfoLoaded = false;
```
YoEight, line 35: Why it's needed to know if we have already loaded the server info or not?

YoEight (dismissed):

db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
shaan1337: based on my tests, it appears that the version is not parsed correctly (split() takes in a regex, not a delimiter)

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
shaan1337: will fix, thanks

db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
shaan1337: @YoEight ok, does this imply changing the signature to return `Optional<ServerVersion>` instead of `ServerVersion`?
```
public Optional<ServerVersion> getServerVersion()
```

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
shaan1337: this is for the case where the client hasn't connected to any node yet, so no server version is available/known yet. i'm not sure if that's the best way to handle it

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
shaan1337: this is to detect if we've already connected to at least one node and fetched the server info in order to make sure that when getServerVersion() is called, we either return a correct value (which may be slightly stale if a race condition occurs where the client connects to another node) or throw an exception if it's not populated yet

db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
YoEight: But if it were the case, it would mean it never worked before.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
YoEight: That shouldn't happen. If it does it now, it means some regression was introduced with recent PRs.

db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
YoEight: I would prefer yes.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
YoEight: See: https://github.com/EventStore/EventStoreDB-Client-Java/pull/165#discussion_r930520577

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
shaan1337: say somebody does the following:
```
SingleNodeClient client = new SingleNodeClient(...);
client.getServerVersion();
```

given that messageLoop() is executed asynchronously in the constructor, isn't there a time window where it's possible that the server info isn't yet loaded?

db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
shaan1337: yes, it appears to be the case but it wasn't noticeable since the version wasn't being used anywhere yet.

shaan1337: e.g "22.6.0".split(".") returns a 0-length array, so major, minor and patch were not set and remained 0.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
shaan1337: fixed

db-client-java/src/main/java/com/eventstore/dbclient/ServerVersion.java
shaan1337: fixed.

db-client-java/src/main/java/com/eventstore/dbclient/ServerFeatures.java
YoEight: I see, thanks for the explanation.

db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
YoEight: I would suggest doing as we do with other clients (Go or the Rust clients for example). Meaning getting the server version should be an async operation. I truly believed that was the case since the beginning, but maybe I overlooked some changes made in recent PRs.

If it takes too much time for you to fix it, keep the implementation you did, and I'll change it in another PR. Please create an issue, so I remember.

YoEight (approved):
If you don't have the time to apply https://github.com/EventStore/EventStoreDB-Client-Java/pull/165#discussion_r934051760, don't sweat it, I'll address it in another PR.

Nice work!

pvanbuijtene (approved):

Issue #170, opened 2022-08-09
Prioritize reconnect after connection loss in GrpcClient


dpasek-senacor:
Currently the `GrpcClient` implementation triggers an automatic reconnect if an operation fails due to connection loss. This reconnect is queue inside the `GrpcClient` after all other open work items to be executed.
This means, that already queued work items will most probably fail as the reconnect is not performed until all these work items have been executed against the lost server connection.

A simple solution to this problem would be to prioritize the work item for connecting to the server (`CreateChannel)` over other work items, i.e. always adding it at the beginning of the work item queue instead of the end.

I will provide a PR for this.

PR #174, opened 2022-08-16
Issue-170 Add CreateChannel message at start of work item queue.

dpasek-senacor:
Improvement for https://github.com/EventStore/EventStoreDB-Client-Java/issues/170

Implementation approach:
- The internal work queue is converted into a "deque" which supports adding items at start and end of queue.
- `CreateChannel` commands are automatically always added to the start of the work queue

Benefits:
- In case of connection failover (due to connection loss or not leader) the number of failing operations will be reduced.

Drawbacks:
- Issued `Shutdown` commands will be delayed, i.e. even if the client has been shutdown a reconnect is performed.
Example: client executes an operation, receives an exception that operation failed to connection loss, client code wants to shutdown the client => shutdown will be executed after performing the reconnect attempts (which might take some time).


dpasek-senacor:
I will retract this PR as discussed internally and we will look for a more robust solution regarding a imrpoved connect/reconnect strategy for the GrpcClient.

PR #186, opened 2022-08-26
Shutdown `GossipClient` after usage.

dpasek-senacor:
PR for https://github.com/EventStore/EventStoreDB-Client-Java/issues/185


db-client-java/src/main/java/com/eventstore/dbclient/GossipClient.java
```diff
@@ -25,8 +23,12 @@ public GossipClient(ManagedChannel channel) {
         _stub = GossipGrpc.newStub(_channel);
     }
 
-    public void shutdown() throws InterruptedException {
-        _channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+    public void shutdown() {
```
dpasek-senacor, line 26: The `shutdown()` method already existed but was unused.

`InterruptedException` is handled here instead of being escalated to user.

pvanbuijtene (approved):

hayley-jean (approved):

Issue #190, opened 2022-09-07
Reactive stream reads might call onError twice


dpasek-senacor:
If a read operation of a stream causes an exception the `onError()` method of a reactive subscriber might be called twice. This is not allowed according to the reactive spec.
.
This is not a problem for the non-reactive API as the second `onError()` call is ignored by the `ReadSuscriber` used in the `EventStoreDBClient`. 
When using the reactive API this might be a problem in certain conditions: framework like RxJava will unsubscribe after the first `onError()` call. The second error cannot be handled and will be delegated to the default error handler. For RxJava this will internally usually call the `UncaughtExceptionHandler` of the current thread. Depending of the context this might be ignored but in certain contexts this might break something. In our scenario it have been some tests.

I will provide an pull request for this as everything to fix it is already in place.

PR #193, opened 2022-09-09
Issue-190 Fix condition causing subscribers not to be completed.

dpasek-senacor:
I'm sorry but the PR regarding the double delivery was already merged but we did not see that there was an error. Read operation now did not complete. See tests runs.

https://github.com/EventStore/EventStoreDB-Client-Java/issues/190

Here is the fix for this


timothycoleman (approved):
Ah I see, compareAndSet returns whether it successfully set the value, rather than the the old value 👍

hayley-jean (approved):

PR #195, opened 2022-10-14
CI test for QA 22.10

shaan1337:


shaan1337:
re-running after building testdata container image

PR #215, opened 2023-03-01
fix(EventDataBuilder): divided JSON and binary EventDataBinary

dbakhmatskyiST:
https://github.com/EventStore/EventStoreDB-Client-Java/issues/216

When we trying to create a new 'EventDate' using 'EventDataBuilder' we have new exception, because in this builder hardcoded 'JsonMapper'.


YoEight:
Thanks for you contribution but I'm closing this PR in favor of #217

PR #226, opened 2023-03-17
Refactor internal connection code.

### LINKED ISSUE #223: Improve InterruptedException handling in com.eventstore.dbclient.GrpcClient.
Current implementation of `com.eventstore.dbclient.GrpcClient.pushMsg(Msg)` logs and swallows the `InterruptedException`. It seems more appropriate to propagate the exception instead (perhaps wrapped in `RuntimeException` or similar), since this is called on application/caller thread.

There might be other places (in `GrpcClient` or elsewhere) which could benefit from stricter interrupt handling. (I have checked `GrpcClient` code but do not have sufficient understanding of client lifecycle to detect any further problems).

YoEight:
Changed: Improve internal gRPC connection management for better error propagation.

Fixes #223

Major rewrite of how gRPC connections are managed internally. This new design allows exceptions like `InterruptedException` to be propagated even when they occur in the connection manager thread.

The public client method `shutdown` changed. It used to be `void` but now it's `CompletableFuture<Void>`. Technically, it shouldn't break user code unless they have enabled very pedantic compiler options.

However I deprecated the `Endpoint` type as using `InetSocketAddress` across the board was simpler and removed unnecessary allocations when converting `Endpoint` to `InetSocketAddress`.

Technically those breaking changes, while minor, warrant upping the major release number.


.github/workflows/pull-requests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         version: [ 20.10.5-buster-slim, 21.10.1-buster-slim, ci ]
+         version: [ previous-lts, lts, latest ]
```
yehohanan7: why did this change? my PR also had some trouble running tests on 20.10.5-buster-slim, may be it's related to your fix...

YoEight: I don't know why `20.10.5-buster-slim` stopped working all of the sudden. However, on the other PR I moved to `previous-lts`, `lts` and `latest` because that's what I should have done a while ago. I didn't know we pulled the trigger on the new naming.

pvanbuijtene:
Before merging, let's discuss how this should be mentioned in the release notes. /cc @ylorph

.github/workflows/pull-requests.yml
pvanbuijtene: Can we keep the `ci`? should be fixed now.

pvanbuijtene (approved):

.github/workflows/pull-requests.yml
YoEight: Done

tambeau:
Request review from Claude as we would like someone strong in Java to look at this.

cdevarenne (approved):
Looks good to me, I am new to the code base (this is day 2 for me).

Issue #225, opened 2023-03-17
Implement User Management for Java client


hayley-jean:
Create a new `EventStoreDBUsersClient` which supports the following operations:
- [x] CreateUser
- [x] GetUser
- [x] DeleteUser
- [x] EnableUser
- [x] DisableUser
- [x] UpdateUser
- [x] ChangeUserPassword
- [x] ResetUserPassword

YoEight:
https://github.com/EventStore/EventStoreDB-Client-Java/pull/227#issuecomment-1578577560

PR #227, opened 2023-03-20
Added: APIs for user management

yehohanan7:
Added: user management APIs (ticket [here](https://github.com/EventStore/EventStoreDB-Client-Java/issues/225))

The new APIs favours the return of `CompletableFuture<Void>` for side-effect only APIs to convey the intent clearly, in order to get that working the `GrpcClient` has to be changed to handle completion of the future with a `null` value.

Updated the integration tests to be able to start a test container in secure mode with test certificates.


db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
```diff
@@ -118,9 +114,14 @@ public <A> CompletableFuture<A> runWithArgs(Function<WorkItemArgs, CompletableFu
...
-                logger.debug("RunWorkItem[{}] completed exceptionally: {}", msgId, error.toString());
 
-                result.completeExceptionally(error);
+                if (error != null) {
+                    logger.debug("RunWorkItem[{}] completed exceptionally: {}", msgId, error.toString());
+                    result.completeExceptionally(error);
+                    return;
+                }
+
+                result.complete(outcome);
```
yehohanan7: The old code was not allowing us to complete a future with `null` value

YoEight: Good catch!

YoEight (approved):

hayley-jean:
Closing this for now, as we may want to put user management into a dedicated tool rather than including it in each client

PR #230, opened 2023-04-11
Implement `ExpectedRevision` raw long representation.

### LINKED ISSUE #228: Next expected version no longer available in version 4+
Hi,

We are currently at client version 3.0.2.

When performing an appendToStream, we get a write result back which contains a nextExpectedVersion, which is a long. The caller then receives this version, and we regularly use it to poll the read model to check when it’s available.

I have been experimenting a bit with client version 4.1.1 to see what changes need to be made. I noticed there isn’t a nextExpectedVersion anymore.

Instead there’s a nextExpectedRevision, but this is an abstract class of ExpectedRevision. One of the subclasses is a SpecificExpectedRevision, but that isn’t public, and so we can’t extract the version from that.

Would it be possible to expose the revision number?

YoEight:
Added: Implement `ExpectedRevision` raw long representation.

Fixes #228

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ExpectedRevision.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ExpectedRevision.java
@@ -43,13 +43,16 @@ public abstract class ExpectedRevision {
         return new SpecificExpectedRevision(revision);
     }
 
-    static ExpectedRevision fromRaw(long revision) {
+    public static ExpectedRevision fromRawLong(long revision) {
         if (revision == -1)
             return ExpectedRevision.noStream();
         if (revision == -2)
             return ExpectedRevision.any();
         if (revision == -4)
             return ExpectedRevision.streamExists();
+
+        if (revision < 0)
+            throw new RuntimeException(String.format("Invalid expected revision long representation '%s'", revision));
 
         return ExpectedRevision.expectedRevision(revision);
     }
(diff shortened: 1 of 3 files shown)


Lougarou (approved):

db-client-java/src/main/java/com/eventstore/dbclient/ExpectedRevision.java
```diff
@@ -43,14 +43,17 @@ public static ExpectedRevision expectedRevision(long revision) {
...
+    public static ExpectedRevision fromRawLong(long revision) {
         if (revision == -1)
             return ExpectedRevision.noStream();
         if (revision == -2)
             return ExpectedRevision.any();
         if (revision == -4)
             return ExpectedRevision.streamExists();
 
+        if (revision < 0)
+            throw new RuntimeException(String.format("Invalid expected revision long representation '%s'", revision));
```
Lougarou, line 55: Side question not 100% related to this PR but how do you actually catch this? I always get another exception for example an unknown exception: 
'Exception in thread "main" java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: UNKNOWN: Exception was thrown by handler'

using:
 AppendToStreamOptions appendToStreamOptions = AppendToStreamOptions.get()
                .expectedRevision(ExpectedRevision.fromRawLong(-3));
        WriteResult writeResult = eventStore
                .appendToStream("accounts", appendToStreamOptions, event)
                .get();
                
update nevermind will check again after: https://github.com/EventStore/EventStoreDB-Client-Java/pull/226

YoEight: Just in case, there is a test in that PR and you could see that I catch the `RuntimeException` directly. I really don't see what you situation could have been.

hayley-jean (approved):

Issue #237, opened 2023-08-27
RecordedEvent misses "isJson" field


michael-schnell:
System metadata has a field called "is-json" which currently is not included in "RecordEvent" as all other meta data that has a key in "SystemMetadataKeys". It should also be included in the resulting record event.

michael-schnell:
Please ignore first (already closed) pull request, it was a mistake. The second one is the right one.

michael-schnell:
Alternative according to suggestion of @YoEight:
> if json then content-type is application/json otherwise it's binary (application/octet-stream)

PR #239, opened 2023-08-27
Added missing "is-json" data.

michael-schnell:
Solves issue #237


YoEight:
Hey @michael-schnell, the `is-json` key is not used. However you could derive the `isJson` property out of the content type property.

if json then content-type is `application/json` otherwise it's binary (`application/octet-stream`)

PR #251, opened 2023-11-02
Allow the user to use whatever gRPC implementation they want.

### LINKED ISSUE #249: ConnectionState.connect() should use generic Grpc.newChannelBuilderForAddress()
EventStoreDB-Client Version: 5.1.0

The [`ConnectionState.connect(...)` method](https://github.com/EventStore/EventStoreDB-Client-Java/blob/1e44f3870c59599977909f467a556f0faac27308/db-client-java/src/main/java/com/eventstore/dbclient/ConnectionState.java#L63) currently creates a managed gRPC channel using the `NettyChannelBuilder` from `io.grpc:grpc-netty-shaded`. We have a project that uses gRPC for multiple things and needs to use the `io.grpc:grpc-netty` transport instead of the shaded one. Because the connect method uses the netty shaded channel builder, this is impossible.

My suggested change is to move to using one of the flavors of `Grpc.newChannelBuilder()` function, allowing gRPC to dynamically choose the correct transport based on what is available on the classpath.

YoEight:
Changed: Allow the user to use whatever gRPC implementation they want.

Fixes #249 

This patch allows the user to use whatever gRPC implementation they want which is valuable if you want to use the client in Android for example. As a consequence, users will have to import `grpc-netty`, `grpc-netty-shaded` or `grpc-okhttp` depending on their situation.

### DIFF
--- a/db-client-java/build.gradle
+++ b/db-client-java/build.gradle
@@ -42,12 +42,12 @@ dependencies {
     implementation "com.google.protobuf:protobuf-java:${protobufVersion}"
     implementation "com.google.protobuf:protobuf-java-util:${protobufVersion}"
     implementation 'com.google.errorprone:error_prone_annotations:2.23.0'
-//    implementation "io.grpc:grpc-netty-shaded:${grpcVersion}"
     implementation "io.grpc:grpc-stub:${grpcVersion}"
     implementation "io.grpc:grpc-protobuf:${grpcVersion}"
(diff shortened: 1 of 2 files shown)


YoEight:
I would put that PR on hold because I'm unable to disable certificate verification using `Grpc.newChannelBuilderForAddress` and `TlsChannelCredentials` only. Setting a custom insecure trust manager and a hostname verifier do absolutely nothing. Will investigate what the folks working on `grpc-netty-shaded` did to make that possible.

PR #258, opened 2023-12-11
Expose building a JSON payload event with raw bytes

### LINKED ISSUE #257: Expose building a JSON payload event with raw bytes in the EventData class
Hey @OliverJurkschat, 
    
    Something is missing indeed but from our side, an oversight to be precise. We deprecated relying on JsonMapper on any public API, like the EventDataBuilder you mentioned. 
    
    The appropriate method is available if you look at `EventDataBuilder` class directly: https://eventstore.github.io/EventStoreDB-Client-Java/com/eventstore/dbclient/EventDataBuilder.html#json(java.lang.String,byte%5B%5D)
    
    With this method, because you can provide the raw bytes yourself, you can use what ever you like to build your JSON. We forgot to expose that method in the `EventData` class.

_Originally posted by @YoEight in https://github.com/EventStore/EventStoreDB-Client-Java/issues/256#issuecomment-1843067390_

w1am:
Added: Expose building a JSON payload event with raw bytes

Fixes: #257

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/EventData.java
@@ -62,6 +62,7 @@ public final class EventData {
      * @return an event data builder.
      * @param <A> a type that can be serialized in JSON.
      */
+    @Deprecated
     public static <A> EventDataBuilder builderAsJson(String eventType, A eventData) {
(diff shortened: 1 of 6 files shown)


YoEight (approved):
Nicely done!

cdevarenne (approved):

PR #288, opened 2024-10-06
Add support for direct DNS lookups behind a feature flag.

### LINKED ISSUE #273: Should DNS discovery use all IPs in a multi-address DNS name as cluster seeds?
My experience is that DNS discovery fails when 1 node out of 3 is down, and discovery spends all of `maxDiscoverAttempts` trying to get gossip from the node that is down, instead of also considering the other 2 nodes' IPs registered with a multi-address DNS name.

I was able to implement the behaviour I expect like this

```java
# ClusterDiscovery.java:

    void discover(ConnectionState state) {
-       List<InetSocketAddress> candidates = new ArrayList<>(this.seeds);
+       List<InetSocketAddress> candidates = new ArrayList<>();
+
+       if (state.getSettings().isDnsDiscover()) {
+           try {
+               InetSocketAddress dnsSeed = this.seeds.get(0);
+
+               // Resolve cluster DNS name
+               candidates = Arrays.stream(InetAddress.getAllByName(dnsSeed.getHostName()))
+                   .map(addr -> new InetSocketAddress(addr, dnsSeed.getPort()))
+                   .collect(Collectors.toList());
+                  
+           } catch (UnknownHostException e) {
+               throw new UncheckedIOException(e);
+           }
+       } else {
+           candidates = new ArrayList<>(this.seeds);
+       }

        if (candidates.size() > 1) {
```
I'm not sure, however, if you'd prefer to delegate somehow this behaviour to the gRPC client, since it's the gRPC client that currently does the lookup.

YoEight:
Added: Add support for direct DNS lookups behind a feature flag.

Fixes #273 

### description

With this feature, the client will directly perform DNS lookups when probing for candidates during the connection process to a cluster of nodes. This mirrors the behavior from when the TCP API was still supported.

You can enable that feature using a connection string or a the connection builder.

### connection string

```
esdb+discover://my.esdbcluster.com:1234?feature=dns-lookup
```

### connection builder

```java
     EventStoreDBClientSettings.builder()
        .dnsDiscover(true)
        .addHost("my.esdbcluster.com", 1234)
        .feature("dns-lookup")
        .buildConnectionSettings()
```

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ClientFeatureFlags.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ClientFeatureFlags.java
@@ -1,0 +1,10 @@
+package com.eventstore.dbclient;
+
+public final class ClientFeatureFlags {
+    /**
+     * Enables direct DNS name resolution, retrieving all IP addresses associated with a given hostname. This
+     * functionality was initially implemented to support the now-deprecated TCP API. It is particularly useful in
+     * scenarios involving clusters, where node discovery is enabled.
+     */
+    public static final String DNS_LOOKUP = "dns-lookup";
+}
--- a/db-client-java/src/main/java/com/eventstore/dbclient/ClusterDiscovery.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/ClusterDiscovery.java
@@ -1,5 +1,9 @@
 package com.eventstore.dbclient;
 
+import com.eventstore.dbclient.resolution.DeferredNodeResolution;
+import com.eventstore.dbclient.resolution.DeprecatedNodeResolution;
+import com.eventstore.dbclient.resolution.FixedSeedsNodeResolution;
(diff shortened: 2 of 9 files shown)


db-client-java/src/main/java/com/eventstore/dbclient/ClientFeatureFlags.java
```diff
@@ -0,0 +1,10 @@
+package com.eventstore.dbclient;
+
+public final class ClientFeatureFlags {
+    /**
+     * Enables direct DNS name resolution, retrieving all IP addresses associated with a given hostname. This
+     * functionality was initially implemented to support the now-deprecated TCP API. It is particularly useful in
+     * scenarios involving clusters, where node discovery is enabled.
+     */
+    public static final String DnsLookup = "dns-lookup";
```
lbodor: case: `dnsLookup`

w1am (approved):

PR #290, opened 2024-10-28
Do inject trace context when opentelemetry is disabled

w1am:
Fixed: Do inject trace context when opentelemetry is disabled


w1am:
Currently, the OpenTelemetry library does not offer an easy way to do this

PR #298, opened 2025-01-10
Unwrap CompletionException in GrpcClient

### LINKED ISSUE #297: Exceptions are not properly handled in GrpcClient
Hi,

We have had some issues with `NotLeaderException` and `StatusRuntimeException` (UNAVAILABLE) resulting in a permanent disconnect that is only solved by a restart of the disconnected client.

This happens consistently whenever a leader election changes which node in our cluster acts as the leader.

From what I can see the issue seems to be isolated to appending to streams, as persistent subscriptions (as far as I am currently aware) seem to reconnect. This might not be the case, but I have not really looked into that part further at this moment.

I have managed to reproduce this with debug logs, as well as locally on my machine, by connecting to a cluster and then killing the leader and attempting to write to a stream.

Whilst debugging the code I think I have identified the issue, which is also apparent if you look at this debug log:
```
RunWorkItem[baeaec4e-165e-4d9f-bc83-0f4b9e60b42c] completed exceptionally: java.util.concurrent.CompletionException: com.eventstore.dbclient.NotLeaderException
```

My guess is that the exception produced by `com.eventstore.dbclient.ClientTelemetry#traceAppend` is not handled in `com.eventstore.dbclient.GrpcClient#runWithArgs` becuase GrpcClient expects a `NotLeaderException` or `StatusRuntimeException` but receives a `CompletionException`.

So if I am correct, this should be solve:able by either not wrapping the exception in ClientTelemetry or by unwrapping CompletionExceptions in the GrpcClient.

Something as simple as:
```Java
// com/eventstore/dbclient/GrpcClient.java:83
if (error instanceof CompletionException) {
    error = error.getCause();
}
```
Should theoretically solve this, at least for this scenario.

Another potential solution would be to not wrap the exception at all, but it would require more intrusive changes to the Future chain. For example:
```Java
// com/eventstore/dbclient/ClientTelemetry.java:108
.exceptionallyCompose(throwable -> {
    span.setStatus(StatusCode.ERROR);
    span.recordException(throwable);
    span.end();
    return CompletableFuture.failedFuture(throwable);
}).thenApply(writeResult -> {
    span.setStatus(StatusCode.OK);
    span.end();
    return writeResult;
});
```

I could provide a PR for this if you agree with me that it should be changed.

If it is of interest I can provide a small (but dirty) unit test that kinda reproduces this, but as I mentioned above - killing the leader node and then attempting a appendToStream should throw this error.

marcuslyth:
GrpcClient expects Exceptions of a specific type but as it is right now it will not handle any exceptions as they are propagated as CompletionException

This will not solve any similar issues for any other class using ClientTelemetry, and a better solution would probably be to never wrap the exception, however that would require more intrusive changes to the future chain.

Fixes #297

### DIFF
--- a/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
+++ b/db-client-java/src/main/java/com/eventstore/dbclient/GrpcClient.java
@@ -9,6 +9,7 @@ import org.slf4j.LoggerFactory;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
@@ -78,6 +79,10 @@ class GrpcClient {
         return getWorkItemArgs().thenComposeAsync((args) -> {
             return action.apply(args).handleAsync((outcome, e) -> {
                 if (e != null) {
+                    if (e instanceof CompletionException) {
+                        e = e.getCause();
+                    }
+
                     if (e instanceof NotLeaderException) {
                         NotLeaderException ex = (NotLeaderException) e;
                         // TODO - Currently we don't retry on not leader exception but we might consider


YoEight:
Hi @marcuslyth, thank you for your contribution!

Great catch. That said, I noticed that your patch will conflict with #295. To ensure you remain credited as the original author of the fix, I’d prefer not to resolve it directly in #295. Once #295 is merged, would you mind rebasing your work onto the trunk branch?

Thanks again for your time and effort!

marcuslyth:
> Hi @marcuslyth, thank you for your contribution!
> 
> Great catch. That said, I noticed that your patch will conflict with #295. To ensure you remain credited as the original author of the fix, I’d prefer not to resolve it directly in #295. Once #295 is merged, would you mind rebasing your work onto the trunk branch?
> 
> Thanks again for your time and effort!

You are more than welcome!

Sure thing, as soon as #295 have been merged I can rebase my change.

YoEight:
Hey @marcuslyth, #295 is finally merged. Could you rebase your branch?

marcuslyth:
@YoEight 

Done!

YoEight:
After all you rebased from master, did you patch is still behaving properly?

marcuslyth:
@YoEight 
> After all you rebased from master, did you patch is still behaving properly?

Yes, I have tested it both on unit level and by building the client and using it in one of our services with a three node setup and it works as expected.

YoEight (approved):

YoEight:
Thanks for your work!

PR #319, opened 2025-04-15
docs: correct article usage with "Kurrent" 

teamchong:
This PR addresses grammatical issues that occurred during the rebranding from "Event Store" to "Kurrent". The main issue was incorrect article usage - using "an KurrentDB" instead of "a KurrentDB" since "Kurrent" starts with a consonant sound.


w1am:
Fixed in https://github.com/kurrent-io/KurrentDB-Client-Java/pull/318

